PR #16, opened 2018-05-21
testing(pipelines): Adds missing pipeline unit tests.

jtk54:
Also fixes a few small bugs found in writing the tests :)

### DIFF
--- a/command/pipelines/pipeline_list_test.go
+++ b/command/pipelines/pipeline_list_test.go
@@ -1,0 +1,153 @@
+package pipelines
+
+import (
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/spinnaker/spin/command"
+)
+
+func TestPipelineList_basic(t *testing.T) {
+	ts := testGatePipelineListSuccess()
+	defer ts.Close()
+
+	meta := command.ApiMeta{}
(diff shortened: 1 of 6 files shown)


command/pipelines/pipeline_list_test.go
```diff
@@ -0,0 +1,155 @@
...
+	ts := testGatePipelineListSuccess()
+	defer ts.Close()
+
+	meta := command.ApiMeta{}
+	args := []string{"--gate-endpoint", ts.URL} // Missing application.
+	cmd := PipelineListCommand{
+		ApiMeta: meta,
+	}
+	ret := cmd.Run(args)
+	if ret == 0 { // Success is actually failure here.
```
lwander: Can you explain why?

lwander: Just asking because in the other tests where you do it "malformed" or "fail" is in the title and that makes it a little clearer.

command/pipelines/pipeline_list_test.go
```diff
@@ -0,0 +1,155 @@
...
+        "requisiteStageRefIds": [],
+        "type": "wait",
+        "waitTime": 30
+      }
+    ],
+    "triggers": [],
+    "updateTs": "1526578883109"
+  }
+]
+
```
lwander: Extra newline?

PR #19, opened 2018-05-23
feat(pipelines): Adds pipeline execute.

jtk54:


lwander:
Does a user have to supply an entire pipeline definition to be executed? This feels a little heavy duty -- I'm thinking a user will usually want to say "run this pipeline already stored in spinnaker".

jtk54:
**Don't Merge**, the pattern of submitting the whole pipeline as the payload isn't well-supported from the UI. We should support executing pipelines by reference only.

jtk54:
Ok PTAL now, fixed this up to execute pipelines like the UI does.

command/pipelines/pipeline_execute.go
```diff
@@ -0,0 +1,99 @@
...
+	name        string
+}
+
+// flagSet adds all options for this command to the flagset and returns the
+// flagset object for further modification by subcommands.
+func (c *PipelineExecuteCommand) flagSet() *flag.FlagSet {
+	cmd := "pipeline execute"
+
+	f := c.ApiMeta.GlobalFlagSet(cmd)
+	f.StringVar(&c.application, "application", "", "Spinnaker application the pipeline lives in")
```
lwander: Are parameters & artifacts being handled in a separate PR?

command/pipelines/pipeline_execute.go
```diff
@@ -0,0 +1,99 @@
...
+	f.Usage = func() {
+		c.ApiMeta.Ui.Error(c.Help())
+	}
+
+	return f
+}
+
+// executePipeline calls the Gate endpoint to execute the pipeline.
+func (c *PipelineExecuteCommand) executePipeline() (gate.HttpEntity, *http.Response, error) {
+	entity, resp, err := c.ApiMeta.GateClient.PipelineControllerApi.InvokePipelineConfigUsingPOST1(nil,
```
lwander: How does this work? I thought we still needed to send a pipeline config

command/pipelines/pipeline_execute.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	_, resp, err := c.executePipeline()
+
+	if err != nil {
+		c.ApiMeta.Ui.Error(fmt.Sprintf("%s\n", err))
+		return 1
+	}
+
+	if resp.StatusCode != http.StatusAccepted {
+		c.ApiMeta.Ui.Error(fmt.Sprintf("Encountered an error saving pipeline, status code: %d\n", resp.StatusCode))
```
lwander: Is this a save?

command/pipelines/pipeline_execute.go
jtk54: `Invoke` searches for a config based on app/name (like `get` does) and starts that config. We _can_ send an explicit config to run but that method may cause the pipeline execution not to be tracked in the UI, which doesn't seem like what we want.

command/pipelines/pipeline_execute.go
jtk54: Yes, I want to get the rudimentary operations done before expanding.

PR #23, opened 2018-05-29
fix(flags): Strip spaces from string lists parsed from flags.

jtk54:


command/applications/application_save.go
```diff
@@ -20,7 +20,13 @@ func (f *flagStringArray) String() string {
 }
 
 func (f *flagStringArray) Set(value string) error {
-	*f = strings.Split(value, ",")
+	stripped := make([]string, 0)
```
lwander: Why only for application_save?

jtk54: Not used for any others at this point, should I just factor these out into a lib at this point?

lwander: Up to you, was just curious what the pattern was

jtk54:
Factored into a `util` library.

PR #24, opened 2018-05-30
feat(applications): Adds application GET operation.

jtk54:


command/applications/application_get.go
```diff
@@ -0,0 +1,96 @@
...
+	applicationName string
+}
+
+// flagSet adds all options for this command to the flagset and returns the
+// flagset object for further modification by subcommands.
+func (c *ApplicationGetCommand) flagSet() *flag.FlagSet {
+	cmd := "application get"
+
+	f := c.ApiMeta.GlobalFlagSet(cmd)
+	f.StringVar(&c.applicationName, "application-name", "", "Name of the application")
```
lwander: Why make this a flag and not a positional argument? If you always expect the name I'd think

`spin application get my-app` 

feels natural

jtk54: SGTM

PR #27, opened 2018-06-14
feat(applications): Adds delete operation for applications.

jtk54:


jtk54:
Also a few small, random string formatting fixes.

PR #28, opened 2018-06-18
docs(readme): Update readme with usage instructions for alpha.

jtk54:


README.md
```diff
@@ -67,3 +67,9 @@ auth:
     keyPath: ~/.spin/keyfile
     certPath: ~/.spin/certfile
 ```
+
+# Usage
+
+Download the `spin` binary from a [release](https://github.com/spinnaker/spin/releases) and
```
lwander: How is the download done? Can we provide an easy curl command?

jtk54: Is there a latest release url for github?

jtk54: I guess we could set a `$VERSION` and then give a curl command.

jtk54:
How about this?

lwander:
Do we have to crosscompile for darwin or linux? Or just say this is linux only?

jtk54:
Linux only for now in the pre-release, I'll mark it there. If we want, I can figure out how to make a cross-compatible release for a proper alpha.

PR #29, opened 2018-06-21
chore(release): add release process

lwander:


lwander:
They are encrypted and stored here: https://github.com/spinnaker/spin/pull/29/files#diff-efd60fa4ac0a995d537f2fcae64c0cdc

PR #33, opened 2018-06-22
chore(pipeline/save): Fix frivolous output.

jtk54:


lwander:
@jtk54 can you rebase?

PR #37, opened 2018-06-24
feat(build): Use dep for dependency management

burdzwastaken:
vgo might also be worth exploring although I haven't looked into it too much but certainly in my local development I find it way more useful to use `dep` than `go get` :D


Gopkg.toml
```diff
@@ -0,0 +1,54 @@
...
+  branch = "master"
+  name = "golang.org/x/net"
+
+[[constraint]]
+  branch = "master"
+  name = "golang.org/x/oauth2"
+
+[[constraint]]
+  name = "gopkg.in/yaml.v2"
+  version = "2.2.1"
```
jtk54: Is there a reason this version needs to be pinned? Seems like `HEAD` for the dependencies is OK for now, especially since this is very early alpha and likely to change. If/when we get to a point where we proliferate a bunch of dependencies and require specific versions of pinning we can add it, but I think this is unnecessary overhead for now.

burdzwastaken: no specific reason this version needs to be pinned at all. completely agree where `HEAD` is fine for a smaller, alpha codebase.

burdzwastaken:
closing for now. can be explored when vendoring becomes more complex.

PR #38, opened 2018-06-27
feat(auth/oauth2): Support for OAuth2 as an authentication method.

jtk54:
Needs docs as well.

### DIFF
--- a/command/meta.go
+++ b/command/meta.go
@@ -15,14 +15,18 @@
 package command
 
 import (
+	"bufio"
+	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/json"
 	"errors"
 	"flag"
 	"fmt"
 	"io/ioutil"
 	"net/http"
 	"net/http/cookiejar"
+	_ "net/http/pprof"
 	"os"
 	"os/user"
 	"path/filepath"
@@ -33,6 +37,7 @@ import (
 	"github.com/mitchellh/go-homedir"
 	"github.com/spinnaker/spin/config"
 	gate "github.com/spinnaker/spin/gateapi"
+	"golang.org/x/oauth2"
 	"gopkg.in/yaml.v2"
 )
 
@@ -50,6 +55,9 @@ type ApiMeta struct {
 
 	// Spin CLI configuration.
 	Config config.Config
+
+	// Context for OAuth2 access token.
+	Context context.Context
 
 	// Internal fields
 	color bool
@@ -122,6 +130,16 @@ func (m *ApiMeta) Process(args []string) ([]string, error) {
 	}
 
 	// Api client initialization.
+	err = m.Authenticate()
+	if err != nil {
+		m.Ui.Error(fmt.Sprintf("OAuth2 Authentication failed."))
+		return args, err
+	}
+	m.Ui.Output(fmt.Sprintf("contex with oauth2 access token: %v\n", m.Context))
+	m.Ui.Output(fmt.Sprintf("just the access token: %s\n", m.Context.Value(gate.ContextAccessToken).(string)))
+	// TODO: token.SetAuthHeader() in the actual calls.
+	// Prolly do this in some type of filter, or in the DefaultHeader, or in the Client.
+
 	client, err := m.InitializeClient()
 	if err != nil {
 		m.Ui.Error(fmt.Sprintf("Could not initialize http client, failing."))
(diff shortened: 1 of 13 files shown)


command/meta.go
```diff
@@ -212,11 +226,96 @@ func (m *ApiMeta) initializeX509Config(client http.Client, clientCA []byte, cert
...
+		usr, err := user.Current()
+		if err != nil {
+			m.Ui.Error("Could not read current user from environment, failing.")
+			return err
+		}
+		credentialsLocation := filepath.Join(usr.HomeDir, ".spin", "credentials")
+		config := &oauth2.Config{
+			ClientID:     OAuth2.ClientId,
+			ClientSecret: OAuth2.ClientSecret,
+			RedirectURL:  "http://localhost:8085",
```
lwander: What does hardcoding this do?

command/meta.go
```diff
@@ -212,11 +226,96 @@ func (m *ApiMeta) initializeX509Config(client http.Client, clientCA []byte, cert
...
+
+			newToken, err = config.Exchange(oauth2.NoContext, code)
+			if err != nil {
+				return err
+			}
+		}
+
+		m.Ui.Output("Caching oauth2 token.")
+		buf, _ := json.Marshal(&newToken)
+		ioutil.WriteFile(credentialsLocation, buf, 0777) // TODO: Which file mode?
```
lwander: Can you lookup the filemode above? I probably would recommend against merging this until we have that figured out, since this makes the creds globally readable which isn't great.

command/meta.go
```diff
@@ -212,11 +226,96 @@ func (m *ApiMeta) initializeX509Config(client http.Client, clientCA []byte, cert
...
+			http.Handle("/", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				code := r.FormValue("code")
+				fmt.Fprintln(w, code)
+			}))
+			go http.ListenAndServe(":8085", nil)
+			// Note: leaving server connection open for scope of request, will be reaped on exit.
+
+			authURL := config.AuthCodeURL("state-token", oauth2.AccessTypeOffline, oauth2.ApprovalForce)
+			m.Ui.Output(fmt.Sprintf("Navigate to %s and authenticate", authURL))
+			code := m.Prompt()
```
lwander: Does this mean this can only be run in interactive mode?

command/meta.go
```diff
@@ -212,11 +226,96 @@ func (m *ApiMeta) initializeX509Config(client http.Client, clientCA []byte, cert
...
+		if !OAuth2.IsValid() {
+			return errors.New("Incorrect OAuth2 auth configuration.")
+		}
+
+		usr, err := user.Current()
+		if err != nil {
+			m.Ui.Error("Could not read current user from environment, failing.")
+			return err
+		}
+		credentialsLocation := filepath.Join(usr.HomeDir, ".spin", "credentials")
```
lwander: Why not write the token into the ~/.spin/config file?

command/meta.go
jtk54: Hardcodes the address of the local http server we use to catch the `code` response.  This is necessary because you have to whitelist redirect URLs when creating an OAuth2 client in your provider: https://www.spinnaker.io/setup/security/authentication/oauth/providers/google/.

command/meta.go
jtk54: This is in the case where there are no cached credentials. Ideally an operator would run through this path once to cache the credentials. `spin` would use/refresh the cached credentials each time after. I'll call this out in the docs.

command/meta.go
jtk54: Will experiment for the correct filemode.

command/meta.go
jtk54: This is also consistent with what `gcloud` does for instance.

command/meta.go
jtk54: No reason I guess. I imagined users possibly wanting to delete `credentials` to periodically refresh, but there's no reason they couldn't edit the config file either. The stored token is json for the refresh token and expiry information, so I wonder if that will work nicely with yaml - I'll give it a shot.

jtk54:
P I N G

command/meta.go
```diff
@@ -212,11 +228,77 @@ func (m *ApiMeta) initializeX509Config(client http.Client, clientCA []byte, cert
...
 
 	return &client
 }
 
+func (m *ApiMeta) Authenticate() error {
+	auth := m.Config.Auth
+	if auth != nil && auth.Enabled && auth.OAuth2 != nil {
+		OAuth2 := auth.OAuth2
+		if !OAuth2.IsValid() {
+			return errors.New("Incorrect OAuth2 auth configuration.")
```
lwander: Can we add a todo to improve this error?

PR #40, opened 2018-07-03
feat(pipelines): Emit execution Id on pipeline execute command.

jtk54:


command/pipelines/pipeline_execute.go
```diff
@@ -91,7 +113,27 @@ func (c *PipelineExecuteCommand) Run(args []string) int {
...
+	if err != nil {
+		c.ApiMeta.Ui.Error(fmt.Sprintf("%s\n", err))
+		return 1
+	}
+	if resp.StatusCode != http.StatusOK {
+		c.ApiMeta.Ui.Error(fmt.Sprintf("Encountered an error querying pipeline execution, status code: %d\n", resp.StatusCode))
+		return 1
+	}
+	if len(executions) != 1 {
+		c.ApiMeta.Ui.Error(fmt.Sprintf("Expected one forked execution, found %d with refIds: %v.", len(executions), executions))
```
lwander: Where does the term "forked" come from?

command/pipelines/pipeline_execute.go
```diff
@@ -91,7 +113,27 @@ func (c *PipelineExecuteCommand) Run(args []string) int {
...
+	attempts := 0
+	for len(executions) != 1 && attempts < 5 {
+		executions, resp, err = c.queryExecution()
+		attempts += 1
+	}
+	if err != nil {
+		c.ApiMeta.Ui.Error(fmt.Sprintf("%s\n", err))
+		return 1
+	}
+	if resp.StatusCode != http.StatusOK {
```
lwander: Is it possible we return 2xx but non 200 on success?

command/pipelines/pipeline_execute.go
```diff
@@ -61,6 +61,28 @@ func (c *PipelineExecuteCommand) executePipeline() (gate.HttpEntity, *http.Respo
...
+	if resp != nil && resp.StatusCode != http.StatusOK {
+		c.ApiMeta.Ui.Error(fmt.Sprintf("Encountered an error executing pipeline, status code: %d\n", resp.StatusCode))
+		return nil, resp, err
+	}
+
+	return c.ApiMeta.GateClient.ExecutionsControllerApi.GetLatestExecutionsByConfigIdsUsingGET(
+		c.ApiMeta.Context,
+		config["id"].(string),
+		map[string]interface{}{
+			"limit":    int32(1),
```
lwander: What if parallel executions are enabled and we get the wrong one? Should we use @joonlim's new endpoint?

command/pipelines/pipeline_execute.go
jtk54: I don't know, not attached to it.

command/pipelines/pipeline_execute.go
jtk54: Yep SGTM worth a refactor.

command/pipelines/pipeline_execute.go
jtk54: I don't think so, but I don't think it hurts to check for 2xx here.

jtk54:
Updated. I had to regenerate the client lib, the actual added code is split out in the first and third commits @lwander.

jtk54:
Tests are fixed, PTAL.

command/pipelines/pipeline_execute.go
```diff
@@ -91,7 +101,33 @@ func (c *PipelineExecuteCommand) Run(args []string) int {
...
+	if len(executions) == 0 {
+		c.ApiMeta.Ui.Error(fmt.Sprintf("Unable to start any executions, server response was: %v", resp))
+		return 1
+	}
+
+	refIds := make([]string, 0)
+	for _, execution := range executions {
+		refIds = append(refIds, execution.(map[string]interface{})["id"].(string))
+	}
+	c.ApiMeta.Ui.Output(fmt.Sprintf("Started pipeline execution. Pipeline '%s' has %d executions running with ids: %v\n",
```
lwander: I wonder if we should make this more "programmer friendly" by just emitting the list rather than all the extra text

jtk54: I'll change this to output just the list; however, I think we need further changes adding a `-q` quiet output flag to make this really script-able. For instance alongside the list output I have `Caching oauth2 token.` We could probably remove that specific output, but it may make sense to add levels/output suppression.

PR #44, opened 2018-07-17
feat(pipelines): Accept Stdin for pipeline save.

jtk54:


command/pipelines/pipeline_save_test.go
```diff
@@ -46,6 +46,34 @@ func TestPipelineSave_basic(t *testing.T) {
...
+	defer os.Remove(tempFile.Name())
+
+	// Prepare Stdin for test reading.
+	tempFile.Seek(0, 0)
+	oldStdin := os.Stdin
+	defer func() { os.Stdin = oldStdin }()
+	os.Stdin = tempFile
+
+	args := []string{"--gate-endpoint", ts.URL}
+	fmt.Println(args)
```
sbwsg: `*testing.T` has `Log` and `Logf` functions that silence output unless -v flag is passed or the test fails.  Do you intend for this to always be printed as part of test output?

command/pipelines/pipeline_save.go
```diff
@@ -49,11 +51,23 @@ func (c *PipelineSaveCommand) flagSet() *flag.FlagSet {
 
 // parsePipelineFile reads and deserializes the input pipeline file.
```
sbwsg: is this comment out of date now that the func falls back to stdin?  i guess stdin is technically a file.

jtk54: Yep.

command/pipelines/pipeline_save_test.go
jtk54: That shouldn't be there in the first place.

command/pipelines/pipeline_save.go
jtk54: Yep as in it's out of date and I updated the comment*

jtk54:
Closes https://github.com/spinnaker/spinnaker/issues/3058.

PR #45, opened 2018-07-19
chore(release): Update release.sh and TravisCI script.

jtk54:
@lwander FYI.


release.sh
```diff
@@ -3,39 +3,96 @@
...
+        KEY_FILE="$1"
+        shift
+        ;;
+      --version)
+        VERSION="$1"
+        shift
+        ;;
+      --help|-help|-h)
+        print_usage
+        exit 13
```
ewiseblatt: why 13? Doesnt -h usually succeed with 0?

release.sh
```diff
@@ -3,39 +3,96 @@
...
+
+    --version <arg>              Version to tag the `spin` binaries with.
+EOF
+}
+
+
+process_args "$@"
+
+if [ ! -d ${HOME}/google-cloud-sdk ]; then
+  curl https://sdk.cloud.google.com | bash -s --disable-prompts
```
ewiseblatt: Is there a reason you need this rather than any pre-installed gcloud?

release.sh
```diff
@@ -3,39 +3,98 @@
...
+        exit -1
+    esac
+  done
+}
+
+
+function print_usage() {
+    cat <<EOF
+usage: $0 [--gcs_bucket_path <path>]
+    [--key_file <key file path>] [--version <version>]
```
ewiseblatt: key_file and version are required so shouldnt have [] around them. I'd put --gce_bucket_path last since it is optional

release.sh
```diff
@@ -3,39 +3,98 @@
...
+    --version <arg>              Version to tag the `spin` binaries with.
+EOF
+}
+
+
+process_args "$@"
+
+# Google cloud sdk installation from
+# https://cloud.google.com/sdk/docs/downloads-interactive.
+if [ ! -d ${HOME}/google-cloud-sdk ]; then
```
ewiseblatt: why a user-installed gcloud and not the machine default if there is one?

release.sh
```diff
@@ -3,39 +3,98 @@
...
-  echo -e "No version to release specified with the \$VERSION env var, exiting"
+  echo -e "No version to release specified with --version, exiting"
   exit 1
 fi
 
-if [ -z "$GCS_BUCKET_PATH" ]; then
-  echo "No GCS bucket specified using \$GCS_BUCKET_PATH, using gs://spinnaker-artifacts/spin"
-  GCS_BUCKET_PATH="gs://spinnaker-artifacts/spin"
+if [ -z "$SPIN_GCS_BUCKET_PATH" ]; then
+  echo "No GCS bucket specified using \$SPIN_GCS_BUCKET_PATH, using gs://spinnaker-artifacts/spin"
```
ewiseblatt: define the constant at top of file as something like PROD_SPIN_GCS_BUCKET_PATH so it is easier to be aware of.

release.sh
jtk54: Where is that gcloud distribution installed?

release.sh
jtk54: This is mostly to facilitate github releases, which won't have the cloud sdk installed.

release.sh
jtk54: I took this from `InstallSpinnaker.sh`, which had it. Not sure either way. I can change this path to 0.

release.sh
jtk54: Should this just be a `which gcloud` instead?

PR #49, opened 2018-07-28
fix(meta): TrimSpace before returning OAuth2 state token read from STDIN

bernardomk:
@jtk54  This PR fixes #1 by doing TrimSpace to the pasted state token.


bernardomk:
@jtk54 Yeah, don't know what happened, rebase didn't work so I've reset and pushed it again.

jtk54:
Great, LGTM. Thanks again.

PR #50, opened 2018-07-29
feat(endpoint): make gate endpoint configurable

ethanfrogers:
makes the `gateEndpoint` configurable via the config file. the flag
still works if the property is missing.

fixes spinnaker/spinnaker#3094

### DIFF
--- a/command/meta.go
+++ b/command/meta.go
@@ -61,10 +61,7 @@ type ApiMeta struct {
 	// Internal fields
 	color bool
 
-	// This is the set of flags global to the command parser.
-	gateEndpoint string
-
-        ignoreCertErrors bool
+	ignoreCertErrors bool
 
 	// Location of the spin config.
 	configLocation string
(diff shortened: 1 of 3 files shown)


ethanfrogers:
cc @jtk54

command/meta.go
```diff
@@ -75,10 +72,10 @@ type ApiMeta struct {
 func (m *ApiMeta) GlobalFlagSet(cmd string) *flag.FlagSet {
 	f := flag.NewFlagSet(cmd, flag.ContinueOnError)
 
-	f.StringVar(&m.gateEndpoint, "gate-endpoint", "http://localhost:8084",
+	f.StringVar(&m.Config.GateEndpoint, "gate-endpoint", "http://localhost:8084",
```
jtk54: Not sure this is evaluated in the right order, I would expect `--gate-endpoint` to override what is specified in the config.

ethanfrogers:
@jtk54 sorry for the delay, last week was crazy busy for me!

jtk54:
Looks like the `--gate-endpoint` flag isn't working properly, [tests](https://github.com/spinnaker/spin/blob/master/command/applications/application_save_test.go#L41) are breaking. I'd consider keeping the top-level `gateEndpoint` and do the merge manually. I think the flag should override the config property if is it supplied.

ethanfrogers:
@jtk54 i know, i need to find some time to fix it. i'm going to go ahead and close this out until i get a chance to come back to it.

PR #53, opened 2018-08-09
feat(build): Fixing user home bug

blee1170:
There is a issue when running on alpine linux. It will error when trying to find the users home directory:
```
Could not read current user from environment, failing.
user: Current not implemented on linux/amd64
```
This PR will first try to use user.Current, if that fails then it will attempt to get the home directory from the environment.

### DIFF
--- a/command/meta.go
+++ b/command/meta.go
@@ -64,7 +64,7 @@ type ApiMeta struct {
 	// This is the set of flags global to the command parser.
 	gateEndpoint string
 
-        ignoreCertErrors bool
+	ignoreCertErrors bool
 
 	// Location of the spin config.
 	configLocation string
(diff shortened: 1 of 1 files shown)


command/meta.go
```diff
@@ -112,21 +112,30 @@ func (m *ApiMeta) Process(args []string) ([]string, error) {
...
-	m.configLocation = filepath.Join(usr.HomeDir, ".spin", "config")
+	m.configLocation = filepath.Join(userHome, ".spin", "config")
 	yamlFile, err := ioutil.ReadFile(m.configLocation)
 	if err != nil {
 		m.Ui.Warn(fmt.Sprintf("Could not read configuration file from %s.", m.configLocation))
 	}
 
 	if yamlFile != nil {
-		err = yaml.UnmarshalStrict(yamlFile, &m.Config)
+		err = yaml.Unmarshal(yamlFile, &m.Config)
```
jtk54: Is there a reason this changed?

blee1170: For me it was error'ing when I did `go build`. I didn't really intent for that to be part of this PR. I can switch it back if need be.

jtk54: If the parsing was failing, either the config is malformed or spin isn't tolerant enough. In either case it seems unrelated to this PR.

blee1170: Okay. I will update the PR.

jtk54:
Thanks for the contribution :+1:

PR #54, opened 2018-08-13
feat(version): Set and surface CLI version.

jtk54:


lwander:
Do we hard-code the release version, or is it set automatically during a release?

jtk54:
We'd set it during a build/release cycle: https://github.com/spinnaker/spinnaker/pull/3182.

version/version.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"fmt"
+)
+
+// The git commit that was compiled. This is filled in by the compiler.
+var GitCommit string
+
+// Main version number being run right now.
+var Version = "0.4.0"
```
ewiseblatt: Why is version information hardcoded in the source?
It is usually imposed from the outside. If the compiler is filling in the GitCommit,
why cant you also fill in the version (and release phase below)?

version/version.go
```diff
@@ -0,0 +1,41 @@
...
+
+// A pre-release marker for the version. If this is an empty string,
+// then the release is a final release. Otherwise this is a pre-release
+// version e.g. "dev", "alpha", etc.
+var Prerelease = "dev"
```
ewiseblatt: Why not give this a name like ReleasePhase? You can either special case "final" to encode as "" or special case "" to mean final. If you go the encoding route, you can have a function that returns the encoding decorator for the release phase (which would either be "" or "-<phase>" depending on phase being "final" or not)

command/meta.go
```diff
@@ -160,7 +161,7 @@ func (m *ApiMeta) Process(args []string) ([]string, error) {
 	cfg := &gate.Configuration{
 		BasePath:      m.gateEndpoint,
 		DefaultHeader: make(map[string]string),
-		UserAgent:     "Spin CLI version", // TODO(jacobkiefer): Add a reasonable UserAgent.
+		UserAgent:     fmt.Sprintf("%s: %s", version.Header, version.VersionString()),
```
ewiseblatt: sorry, not familiar with programming model here and github UI isnt helpful.
Is the UserAgent the HTTP header? It should be in the format "<program>/<version> ..."
I'm not sure what version.Header is here. The name "header" is odd.

version/version.go
jtk54: We'd set it during a build/release cycle: spinnaker/spinnaker#3182 (look for `ldflags`). It's set here also so we have a reasonable user agent and version for built versions that haven't been released yet.

version/version.go
jtk54: `ReleasePhase` makes sense.

command/meta.go
jtk54: Not tied to the name. Is the "/..." common? Something like `/Spin CLI: <version>`?

jtk54:
Unset the hardcoded version, added a script to support building from source if folks want to do that.

build.sh
```diff
@@ -0,0 +1,83 @@
...
+
+if [ ! -z "$go_os"]; then
+    export $GO_OS=$go_os
+fi
+
+LD_FLAGS="-X github.com/spinnaker/spin/version.Version=${VERSION} \
+-X github.com/spinnaker/spin/version.ReleasePhase=${RELEASE_PHASE}"
+go clean
+go get -d -v
+env CGO_ENABLED=0 go build -ldflags "${LD_FLAGS}" .
```
brandonnelson3: I think "env" is unneccessary.

version/version.go
```diff
@@ -0,0 +1,38 @@
...
+// A pre-release marker for the version. If this is an empty string,
+// then the release is a final release. Otherwise this is a pre-release
+// version e.g. "dev", "alpha", etc.
+var ReleasePhase = "dev"
+
+// User Agent name set in requests.
+const UserAgent = "Spin-CLI"
+
+// VersionString prints the version of the Spin CLI.
+func VersionString() string {
```
brandonnelson3: Version is redundant with the package, consider just String(), so external usage reads version.String().

PR #55, opened 2018-08-21
feat(app): Support saving app from file and stdin.

jtk54:
FYI @MrTrustor.


jtk54:
Closes https://github.com/spinnaker/spinnaker/issues/3215.

PR #56, opened 2018-08-22
feat(apps): Wait on task completion to report success.

jtk54:
Depends on https://github.com/spinnaker/orca/pull/2387.

### DIFF
--- a/command/applications/application_save.go
+++ b/command/applications/application_save.go
@@ -22,6 +22,7 @@ import (
 	"net/http"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/spinnaker/spin/command"
 	"github.com/spinnaker/spin/util"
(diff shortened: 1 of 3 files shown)


command/applications/application_save.go
```diff
@@ -138,6 +139,56 @@ func (c *ApplicationSaveCommand) validApp(app map[string]interface{}) bool {
 	return providers != nil && name != "" && email != ""
 }
 
+// queryTask queries the task for the given ref and returns whether it was
```
lwander: Given the name of this function & this comment I would recommend moving the sleep/retry logic elsewhere. This gives you a clean separation between a success call & a blocking wait until success/fail.

jtk54: SG.

PR #62, opened 2018-09-01
feat(config): gate endpoint configurable

ethanfrogers:
make `gate-endpoint` configurable via config file. the CLI flag takes
precedence over the config file.

### DIFF
--- a/command/meta.go
+++ b/command/meta.go
@@ -81,18 +81,27 @@ type ApiMeta struct {
 func (m *ApiMeta) GlobalFlagSet(cmd string) *flag.FlagSet {
 	f := flag.NewFlagSet(cmd, flag.ContinueOnError)
 
-	f.StringVar(&m.gateEndpoint, "gate-endpoint", "http://localhost:8084",
-		"Gate (API server) endpoint")
+	f.StringVar(&m.gateEndpoint, "gate-endpoint", "",
+		"Gate (API server) endpoint. Default http://localhost:8084.")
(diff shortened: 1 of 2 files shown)


ethanfrogers:
@jtk54 PTAL. i'm not a huge fan of removing the `gate-endpoint` default value but it was the only way to make the precedence work out .

example config
```
gate:
  endpoint: https://my-spinnaker-gate:8084
```

briananderson1222:
@ethanfrogers I was digging around the code and found this (may want to update the `example.config` as well). 

When I try to add:
```
gate:
  endpoint: http://blah.blah:8084
```

I get:
```
Could not deserialize config file with contents: [............................], failing.
yaml: unmarshal errors:
  line 1: field gate not found in type config.Config
```

ethanfrogers:
@briananderson1222 how did you install `spin`? it's likely that a version with this PR hasn't been released yet.

briananderson1222:
@ethanfrogers 

```
curl -LO https://storage.googleapis.com/spinnaker-artifacts/spin/$(curl -s https://storage.googleapis.com/spinnaker-artifacts/spin/latest)/darwin/amd64/spin

chmod +x spin

./spin --version
```
returns: `-dev`

but I noticed that $(curl -s https://storage.googleapis.com/spinnaker-artifacts/spin/latest)
returns: `version-1.1.0`.. which now that i'm typing this does not equal `1.10` :)

PR #63, opened 2018-09-05
feat(meta): Adds version check between spin and gate.

jtk54:
Also adds generated client code for a few missing Gate controllers.


command/meta.go
```diff
@@ -32,13 +32,14 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/hashicorp/go-version"
```
jtk54: Real changes in this file.

command/meta.go
```diff
@@ -171,17 +172,59 @@ func (m *ApiMeta) Process(args []string) ([]string, error) {
...
 
+func (m *ApiMeta) gateVersionCheck() error {
+	spinVersion := spinversion.String()
+	gateVersion, resp, err := m.GateClient.VersionControllerApi.GetVersionUsingGET(m.Context)
+	if err != nil {
+		return err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return errors.New(fmt.Sprintf("Encountered an error getting Gate version, status code: %d\n",
```
sbwsg: You could replace `errors.New(fmt.Sprintf(` with a single `fmt.Errorf(`

command/meta.go
```diff
@@ -171,17 +172,59 @@ func (m *ApiMeta) Process(args []string) ([]string, error) {
...
+		m.Ui.Warn(fmt.Sprintf("Spin version %s does not match Gate version %s. Some API operations may be unsupported.", spinMajMin, gateMajMin))
+	}
+	return nil
+}
+
+func (m *ApiMeta) majorMinor(semVer string) (string, error) {
+	fullSemVer, _ := version.NewVersion(semVer)
+	segments := fullSemVer.Segments()
+	if !(len(segments) >= 3) {
+		return "", errors.New(fmt.Sprintf("Malformed semantic version: %v\n", fullSemVer))
```
sbwsg: Same as above; `fmt.Errorf(`

command/meta.go
```diff
@@ -171,17 +172,59 @@ func (m *ApiMeta) Process(args []string) ([]string, error) {
...
 
+	err = m.gateVersionCheck()
+	if err != nil {
+		return nil, err
+	}
+
 	return args, nil
 }
 
+func (m *ApiMeta) gateVersionCheck() error {
```
sbwsg: I'm curious about the choices to print or not print errors to the UI here. It looks like the only condition where the user will see a warning is a version mismatch; otherwise it looks like this will fail quietly. Is that right?

command/meta.go
```diff
@@ -171,17 +172,59 @@ func (m *ApiMeta) Process(args []string) ([]string, error) {
...
 		DefaultHeader: make(map[string]string),
-		UserAgent:     fmt.Sprintf("%s/%s", version.UserAgent, version.String()),
+		UserAgent:     fmt.Sprintf("%s/%s", spinversion.UserAgent, spinVersion),
 		HTTPClient:    client,
 	}
 	m.GateClient = gate.NewAPIClient(cfg)
 
+	err = m.gateVersionCheck()
+	if err != nil {
+		return nil, err
```
sbwsg: What's the reasoning behind returning `nil, err` vs `args, err` elsewhere in the same function?

command/meta.go
```diff
@@ -171,17 +172,59 @@ func (m *ApiMeta) Process(args []string) ([]string, error) {
...
+	if spinMajMin != gateMajMin {
+		m.Ui.Warn(fmt.Sprintf("Spin version %s does not match Gate version %s. Some API operations may be unsupported.", spinMajMin, gateMajMin))
+	}
+	return nil
+}
+
+func (m *ApiMeta) majorMinor(semVer string) (string, error) {
+	fullSemVer, _ := version.NewVersion(semVer)
+	segments := fullSemVer.Segments()
+	if !(len(segments) >= 3) {
```
sbwsg: Possibly clearer to write `if len(segments) < 3`?

command/meta.go
jtk54: `gateVersionCheck` will warn on a version mismatch, noop if the versions match, and return an error in any other case. This is just adding the initial hook for the version check. Later we will likely either do or suggest an upgrade/downgrade based on Gate's exported version.

jtk54:
**DON'T MERGE.** This change requires a decently large refactor for the tests to remain maintainable.

wjoel:
@jtk54 would it be possible to split out the updating of the generated code to a separate PR? I'd like to contribute support for the pipeline template commands.

jtk54:
I agree on splitting the autogenerated code PRs out. Are the pipeline template controllers missing from master?

wjoel:
Sounds good. Yeah, they were previously not included in the generated Swagger definition, fixed on August 28 in https://github.com/spinnaker/gate/pull/585

jtk54:
This is still tracked [here](https://github.com/spinnaker/spinnaker/issues/3130), but the codebase has substantially changed since this PR was opened. I'll link it in that issue as a reference to implement in the new Cobra command style.

PR #66, opened 2018-09-12
feat(build): add dep and vendor support

sergeylanzman:
- [X] Add dep support for 
- [X] Add vendor folder
- [X] Edit readme and install scripts


jtk54:
Pushing this off until we need proper dependency vendoring support (i.e. folks start hitting diamond dependency problems etc).

PR #70, opened 2018-09-20
fix(cli): delete custom error prints

sergeylanzman:


cmd/pipeline/delete.go
```diff
@@ -80,13 +79,11 @@ func deletePipeline(cmd *cobra.Command, options DeleteOptions) error {
 	resp, err := gateClient.PipelineControllerApi.DeletePipelineUsingDELETE(gateClient.Context, options.application, options.name)
 
 	if err != nil {
-		util.UI.Error(fmt.Sprintf("%s\n", err))
```
jtk54: Are there any calls to `UI.Error` left? Can we remove that function as well?

sergeylanzman: yes, in gateway client and savePipeline

sergeylanzman: UI.Error use in gateway client and savePipeline

PR #73, opened 2018-10-01
chore(gateapi): Update Gate api client.

jtk54:
On the suggestion of: https://github.com/spinnaker/spin/pull/63#issuecomment-425915692.


wjoel:
This is great, thanks!

PR #77, opened 2018-10-01
fix(root): Fix output formatting.

jtk54:


jtk54:
Pushed test fixes as well.

Issue #74, opened 2018-10-01
--version doesn't work


duftler:
I get this at head:
```
$ ./spin --version
Error: unknown flag: --version

unknown flag: --version
```

Works fine at the latest published version:
```
$ spin --version
1.2.0
```

dvehar:
@jtk54 please close this issue.

PR #79, opened 2018-10-15
fix(root): Adds version to support version flags.

jtk54:


jtk54:
Closes https://github.com/spinnaker/spin/issues/74.

PR #86, opened 2018-11-08
fix(oauth2): Fixes API calls using OAuth2.

jtk54:
Gate calls were subtly broken due to OAuth2 redirects.
GET calls were traversing the redirects fine, but POST
calls would be turned into GETs with no body, which
the Gate Controllers fail on. Now, when we resolve the
oauth2 token we immediately establish a session for the life
of the spin command so all calls are authenticated.

### DIFF
--- a/cmd/gateclient/client.go
+++ b/cmd/gateclient/client.go
@@ -55,6 +55,9 @@ type GatewayClient struct {
 	// Generate Gate Api client.
 	*gate.APIClient
 
+	// Raw Http Client to do OAuth2 login.
+	HttpClient *http.Client
+
 	// Spin CLI configuration.
 	Config config.Config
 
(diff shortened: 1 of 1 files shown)


cmd/gateclient/client.go
```diff
@@ -152,15 +155,16 @@ func NewGateClient(flags *pflag.FlagSet) (*GatewayClient, error) {
...
 
 	// Api client initialization.
-	err = gateClient.Authenticate()
+	httpClient, err := gateClient.InitializeClient()
 	if err != nil {
-		util.UI.Error(fmt.Sprintf("OAuth2 Authentication failed."))
+		util.UI.Error(fmt.Sprintf("Could not initialize http client, failing."))
 		return nil, err
 	}
+	gateClient.HttpClient = httpClient
```
ezimanyi: Would it make sense to have `gateClient` set `HttpClient` on itself at the end of `InitializeClient` rather than have the caller do it?  (And possibly just skip returning it at all.)

jtk54: We use the client in two places - one when we do the raw `/login` call and once where we pass to the autogenerated Gate API library.

PR #89, opened 2018-12-06
 feat(pipeline-templates): Adds pipeline templates command and GET.

jtk54:
First commit is the `gateapi` regeneration, second is the substantial `pipeline-templates get` change.


cmd/pipeline-template/pipeline_template.go
```diff
@@ -0,0 +1,43 @@
...
+
+type pipelineTemplateOptions struct{}
+
+var (
+	pipelineTemplateShort   = ""
+	pipelineTemplateLong    = ""
+	pipelineTemplateExample = ""
+)
+
+func NewPipelineCmd(out io.Writer) *cobra.Command {
```
lwander: Why not `newPipelineTemplateCmd`?

jtk54: Good catch, I made a few more changes like that as well.

PR #91, opened 2018-12-10
feat(pipeline-templates): Adds pipeline template list command.

jtk54:


jtk54:
Also fixes a few misnamed variables in tests due to copy-paste errors.

cmd/pipeline-template/list.go
```diff
@@ -0,0 +1,76 @@
...
+		Use:     "list",
+		Aliases: []string{"ls"},
+		Short:   listPipelineTemplateShort,
+		Long:    listPipelineTemplateLong,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return listPipelineTemplate(cmd, options)
+		},
+	}
+
+	options.scopes = cmd.PersistentFlags().StringArrayP("scopes", "", []string{}, "set of scopes to reduce the pipeline template list to")
```
lwander: What are scopes? Do we have docs we can link to here? If not I'd put a TODO in here

jtk54: Added.

PR #94, opened 2018-12-11
fix an issue where we can't read the body of this when passed

nevins-b:
Ran into an issue were the body was being passed as nil. Spinnaker would return a 202 correctly.

```
client, err := gate.NewGateClient(flags)
if err != nil {
    return nil, err
}

raw, err := json.Marshal(tmp)
if err != nil {
    return err
}

resp, err := client.PipelineTemplatesControllerApi.CreateUsingPOST(client.Context, bytes.NewReader(raw))
if err != nil {
    return err
}
```

### DIFF
--- a/gateapi/pipeline_templates_controller_api.go
+++ b/gateapi/pipeline_templates_controller_api.go
@@ -67,7 +67,7 @@ func (a *PipelineTemplatesControllerApiService) CreateUsingPOST(ctx context.Cont
 		localVarHeaderParams["Accept"] = localVarHttpHeaderAccept
 	}
 	// body params
-	localVarPostBody = &pipelineTemplate
+	localVarPostBody = pipelineTemplate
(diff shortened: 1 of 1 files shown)


nevins-b:
Sure thing, I'm working on extending [armory-io/terraform-provider-spinnaker](https://github.com/armory-io/terraform-provider-spinnaker) which is consuming this library directly. The issue that I've seen is that when using it shown above orca failed to create the template because of a NPE when getting the id. This seems to be caused because the post body is empty.

jtk54:
We're using the v2 pipeline templates controller (which is a nearly identical flow) for [saving v2 templates](https://github.com/spinnaker/spin/blob/master/cmd/pipeline-template/save.go#L81) in spin successfully so I'm not sure this will address the issue. 

We've also regenerated this client a few times in recent commits. I'm not sure how you're consuming the `gateapi` but that could explain the drift/breakage as well if you are updating Gate.

nevins-b:
I may be an issue with how we are passing in a Read instead of a json struct, let me do some experimentation.

PR #97, opened 2018-12-13
fix(pipeline): Removes constraint requiring id field.

jtk54:
Helps support pipeline templates and was overly strict anyway
since Orca generates a UUID anyway and the other pipeline commands
operate on name/app pair.


dodizzle:
Will this fix the error when you attempt to upload a pipeline with artifacts?

jtk54:
@dodizzle Unlikely. Is there an tracking issue for that bug?

This allows you to omit the top-level `id` when specifying a pipeline. I don't think that's related to supplying artifacts.

PR #103, opened 2018-12-20
feat(google iap): add iap support for user login

ybai2sc:
add IAP support for auth, add support for Spinnaker instances which has IAP enabled for auth

### DIFF
--- a/cmd/gateclient/client.go
+++ b/cmd/gateclient/client.go
@@ -32,7 +32,7 @@ import (
 	"strings"
 
 	"github.com/spinnaker/spin/config"
-	config2 "github.com/spinnaker/spin/config/auth/iap"
+	iap "github.com/spinnaker/spin/config/auth/iap"
 	"github.com/spinnaker/spin/util"
 	"github.com/spinnaker/spin/version"
 
@@ -359,7 +359,7 @@ func (m *GatewayClient) authenticateOAuth2() error {
 func (m *GatewayClient) authenticateIAP() (string, error) {
 	auth := m.Config.Auth
 	iapConfig := auth.Iap
-	token, err := config2.GetIapToken(*iapConfig)
+	token, err := iap.GetIapToken(*iapConfig)
 	return token, err
 }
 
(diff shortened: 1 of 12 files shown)


cmd/gateclient/client.go
```diff
@@ -31,19 +31,20 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/spinnaker/spin/config"
+	config2 "github.com/spinnaker/spin/config/auth/iap"
```
jtk54: Is there a particular reason this is aliased? Shouldn't this resolve to the `iap` package? If the alias is necessary I'd name it `iap` or something similar.

config/auth/iap/oauth.go
```diff
@@ -0,0 +1,208 @@
...
+func (o *oauthReceiver) killWhenReady(n net.Conn, s http.ConnState) {
+	if o.killMeNow && s.String() == "idle" && o.mainClient == n.RemoteAddr().String() {
+		o.doneChan <- nil
+	}
+}
+
+// NewOAuthConfig creates a new OAuth config
+func (o *oauthReceiver) NewOAuthConfig() *oauth2.Config {
+
+	// return oauth2 Config
```
jtk54: I think this is self-documenting.

config/auth/iap/oauth.go
```diff
@@ -0,0 +1,208 @@
...
+	oauthCode := r.FormValue("code")
+	state := r.FormValue("state")
+	if oauthCode == "" || state == "" {
+		response := fmt.Errorf("Invalid response from Google's Account Service")
+		o.WebOutput(w, "Uh Oh!", response.Error())
+		o.doneChan <- response
+		return
+	}
+
+	var validateStateFunc func(string, string) bool
```
jtk54: Necessary to save this in a local variable?

ybai2sc: thanks, have removed this local variable

config/auth/iap/oauth.go
ybai2sc: thx, removed this comment

cmd/gateclient/client.go
ybai2sc: changed the package name to `iap`

PR #104, opened 2018-12-26
chore(swagger): Added gate-web /projects/* support.

cj-taylor:
Related to https://github.com/spinnaker/spinnaker/issues/3787

### DIFF
--- a/gateapi/api/swagger.yaml
+++ b/gateapi/api/swagger.yaml
@@ -1456,6 +1456,96 @@ paths:
           description: "Forbidden"
         404:
           description: "Not Found"
+  /projects:
+    get:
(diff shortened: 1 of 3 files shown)


jtk54:
Is there a symmetric change in https://github.com/spinnaker/gate for this API update?

cj-taylor:
@jtk54, believe what you're asking for is the inclusion of the `/projects/*` endpoints in the gate-web generated swagger spec. 

If so, here's the PR that was added: https://github.com/spinnaker/gate/pull/690

jtk54:
Yep, thanks for pointing out. Still sifting through emails/notification :D

Issue #105, opened 2019-01-07
Gate client doesn't always return response body


nevins-b:
In many cases it would be very useful to get the response body back, for example `PipelineTemplatesControllerApiService.CreateUsingPOST`. This allows us to monitor the task status for these async calls, without this there is no way to monitor if the template fails to process after being accepted.

jtk54:
Makes sense, I'll add it to the project backlog. If you have a good idea about which ops should return the reference to Orca's async task, by all means feel free to take a stab at implementing this.

nevins-b:
Can do, I think it may be an issue with how swagger-codegen is processing these endpoints. Is there any documentation of how that was run to generate the gateapi?

jtk54:
No docs, but it's a pretty self-contained script: https://github.com/spinnaker/gate/blob/master/swagger/generate_swagger.sh.

jtk54:
The previous script generates the swagger spec, and then we use [swagger-codegen](https://github.com/swagger-api/swagger-codegen) to generate the golang api client in spin.

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

spinnakerbot:
This issue is tagged as 'stale' and hasn't been updated  in 45 days, so we are tagging it as 'to-be-closed'. It will be closed in 45 days unless updates are made. If you want to remove this label, comment:

> @spinnakerbot remove-label to-be-closed

spinnakerbot:
This issue is tagged as 'to-be-closed' and hasn't been updated  in 45 days, so we are closing it. You can always reopen this issue if needed.

PR #106, opened 2019-01-09
fix(gateapi): return request body for pipline-template

nevins-b:
Addresses #105 there are probably other places something similar is needed. I also am unable to get the response body back from the delete call to this controller but there two are a good starting point.

### DIFF
--- a/gateapi/api/swagger.yaml
+++ b/gateapi/api/swagger.yaml
@@ -3779,6 +3779,11 @@ paths:
       responses:
         201:
           description: "Created"
+          schema:
+            type: "object"
+            additionalProperties:
+              type: "object"
+              properties: {}
         202:
           description: "Accepted"
           schema:
@@ -4836,6 +4841,11 @@ paths:
       responses:
         201:
           description: "Created"
+          schema:
+            type: "object"
+            additionalProperties:
+              type: "object"
+              properties: {}
         202:
           description: "Accepted"
           schema:
--- a/gateapi/pipeline_templates_controller_api.go
+++ b/gateapi/pipeline_templates_controller_api.go
@@ -30,13 +30,14 @@ type PipelineTemplatesControllerApiService service
 /* PipelineTemplatesControllerApiService Create a pipeline template.
  * @param ctx context.Context for authentication, logging, tracing, etc.
  @param pipelineTemplate pipelineTemplate
- @return */
-func (a *PipelineTemplatesControllerApiService) CreateUsingPOST(ctx context.Context, pipelineTemplate interface{}) ( *http.Response, error) {
+ @return map[string]interface{}*/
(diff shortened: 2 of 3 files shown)


gateapi/api/swagger.yaml
```diff
@@ -3779,6 +3779,11 @@ paths:
       responses:
         201:
           description: "Created"
+          schema:
```
jtk54: Is there a symmetric gate change for the swagger and gate API change?

gateapi/pipeline_templates_controller_api.go
```diff
@@ -493,19 +500,24 @@ func (a *PipelineTemplatesControllerApiService) UpdateUsingPOST(ctx context.Cont
...
+		return successPayload, localVarHttpResponse, reportError("Status: %v, Body: %s", localVarHttpResponse.Status, bodyBytes)
 	}
 
-	return localVarHttpResponse, err
+	if err = json.NewDecoder(localVarHttpResponse.Body).Decode(&successPayload); err != nil {
+		return successPayload, localVarHttpResponse, err
+	}
+
+
+	return successPayload, localVarHttpResponse, err
```
jtk54: Are there no changes necessary in the commands to output the reference to the Orca op?

jtk54: Ah I see, I meant from `spin`'s commands, e.g. [pipeline-templates save command](https://github.com/spinnaker/spin/blob/master/cmd/pipeline-template/save.go#L107). Are you consuming the [gateapi](https://github.com/spinnaker/spin/tree/master/gateapi) without using `spin`?

nevins-b: Yes, I'm using it as a library. The `spin` command uses the V2 API for this so should be unaffected by these changes.

nevins-b:
The more I look at this the more I feel like this is a bug in `swagger-codegen` since there should be a response body even without these changes according to the swagger we have.

jtk54:
Are these changes manual or were they generated using `swagger-codegen`? If you had to manually add the return bodies, then I agree this is likely an issue either with `swagger-codegen` or with our generation process.

nevins-b:
I manually updated the `swagger.yaml` but the Go code changes where generated by the tool. I'm just surprised that we aren't getting a response body back already since it is defined for the `202 Accepted` response

jtk54:
The manual edits to the swagger.json shouldn't be necessary and ought to be generated through https://github.com/spinnaker/gate/blob/master/swagger/generate_swagger.sh in combination with the swagger rules at https://github.com/spinnaker/gate/blob/master/gate-web/config/gate.yml#L55.

louisjimenez:
Closing this as it has not been updated for some time. Feel free to re-open if there are updates.

PR #110, opened 2019-01-12
feat(windows): Build Windows executable on release

brantburnett:
Also includes .git* changes to assist with development on Windows.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.sh text eol=lf
--- a/release.sh
+++ b/release.sh
@@ -86,7 +86,7 @@ if [ -z "$SPIN_GCS_BUCKET_PATH" ]; then
   SPIN_GCS_BUCKET_PATH=$PROD_SPIN_GCS_BUCKET_PATH
 fi
 
-for elem in darwin,amd64 linux,amd64; do
+for elem in darwin,amd64 linux,amd64 windows,amd64; do
   IFS="," read os arch <<< "${elem}"
   echo "Building for $os $arch"
   env CGO_ENABLED=0 GOOS=$os GOARCH=$arch go build .
--- a/release.sh
+++ b/release.sh
@@ -91,11 +91,16 @@ for elem in darwin,amd64 linux,amd64 windows,amd64; do
   echo "Building for $os $arch"
   env CGO_ENABLED=0 GOOS=$os GOARCH=$arch go build .
 
-  path=${SPIN_GCS_BUCKET_PATH}/${VERSION}/${os}/${arch}/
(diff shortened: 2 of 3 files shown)


.gitattributes
```diff
@@ -0,0 +1 @@
+*.sh text eol=lf
```
jtk54: Is this preserving linux newlines?

jtk54: Just as a warning here, we haven't checked for windows compatibility for much of `spin`'s functionality. The likely set of gotchas is probably present (e.g. file paths).

release.sh
```diff
@@ -86,16 +86,21 @@ if [ -z "$SPIN_GCS_BUCKET_PATH" ]; then
   SPIN_GCS_BUCKET_PATH=$PROD_SPIN_GCS_BUCKET_PATH
 fi
 
-for elem in darwin,amd64 linux,amd64; do
+for elem in darwin,amd64 linux,amd64 windows,amd64; do
```
jtk54: This looks good for this script, but you'd also want to add windows to this list as well to incorporate it into our buildtool: https://github.com/spinnaker/spinnaker/blob/master/dev/buildtool/spin_commands.py#L50.

.gitattributes
brantburnett: Yes, that ensures that Bash scripts keep Linux newlines even when checked out on a Windows machine. By default, text files get Windows newlines when checked out on Windows, which Bash doesn't like if you're using WSL to run Linux on Windows.

Also, I've confirmed basic functionality in Windows, though I haven't tested anything involving file paths yet.

release.sh
brantburnett: Sure thing, what's the difference in when each one is used for builds?

jtk54: This script is executed on a github release, and the other is executed nightly in our Jenkins instance to build the entirety of spinnaker via `buildtool` in spinnaker/spinnaker.

brantburnett: Looking `spin_commands.py`, it seems a bit more involved. It appears to do things like check for the file `spin` in GCS, which would need to be updated to check for `spin.exe` for Windows, etc. Unfortunately, I've never used Python so it might take me a bit to figure it all out.

jtk54: Ah I see, hopefully should be relatively painless. Feel free to take a stab and reach out (me or @ewiseblatt) for help if you get stuck.

brantburnett:
Changes to buildtool are in a new PR: https://github.com/spinnaker/spinnaker/pull/3852

jtk54:
Thanks for the contribution! I'll take a look at the buildtool stuff soon.

PR #114, opened 2019-01-23
fix(app): Fixes specifying apps with flags only.

jtk54:
Previously the command would fail if there was no file input from either a flag or STDIN, which was a regression.

### DIFF
--- a/cmd/application/save.go
+++ b/cmd/application/save.go
@@ -64,9 +64,10 @@ func saveApplication(cmd *cobra.Command, options SaveOptions) error {
 	if err != nil {
 		return err
 	}
-	initialApp, err := util.ParseJsonFromFileOrStdin(options.applicationFile)
+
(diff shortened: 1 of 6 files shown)


cmd/application/save.go
```diff
@@ -65,8 +65,12 @@ func saveApplication(cmd *cobra.Command, options SaveOptions) error {
 		return err
 	}
 	initialApp, err := util.ParseJsonFromFileOrStdin(options.applicationFile)
+	if err != nil && strings.HasPrefix(err.Error(), "No json input") {
```
ezimanyi: I wonder if it would be more robust to pass in a flag to `ParseJsonFromFileOrStdin` controlling whether to error or just return `nil` if there's nothing in the file or stdin.  (Looks like there are a couple of places where we're looking for that particular error string and clearing it out.)

jtk54: Probably makes sense. Maybe even another function.

PR #118, opened 2019-02-04
refactor(app): Factors out orca task polling to its own package.

jtk54:


util/task_watcher.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	return nil
+}
+
+func taskCompleted(task map[string]interface{}) bool {
+	taskStatus, exists := task["status"]
+	if !exists {
+		return false
+	}
+gg
```
ezimanyi: Unintentional?

cmd/orca-tasks/task_watcher.go
```diff
@@ -0,0 +1,92 @@
...
+	id := idFromTaskRef(taskRef)
+	task, resp, err := gateClient.TaskControllerApi.GetTaskUsingGET1(gateClient.Context, id)
+
+	if err != nil {
+		return err
+	}
+	if resp.StatusCode < 200 || resp.StatusCode > 299 {
+		return fmt.Errorf("Encountered an error saving application, status code: %d\n", resp.StatusCode)
+	}
+	if !taskSucceeded(task) {
```
ezimanyi: It looks like if the task is still running, we won't even get to the retry logic...should this be removed entirely and instead fallback to the `taskSucceeded` check after we know the task is complete?

cmd/orca-tasks/task_watcher.go
```diff
@@ -0,0 +1,92 @@
...
+		return fmt.Errorf("Encountered an error saving application, task output was: %v\n", task)
+	}
+
+	attempts := 0
+	for (task == nil || !taskCompleted(task)) && attempts < maxAttempts {
+		id := idFromTaskRef(taskRef)
+		task, resp, err = gateClient.TaskControllerApi.GetTaskUsingGET1(gateClient.Context, id)
```
ezimanyi: Is it deliberate that we don't check `err` or `respStatusCode` here?  (Maybe to just try again if there's a communication error?)

cmd/orca-tasks/task_watcher.go
```diff
@@ -0,0 +1,92 @@
...
+		attempts += 1
+		time.Sleep(time.Duration(attempts*attempts) * time.Second)
```
ezimanyi: This is minor, but we'll always sleep for some amount of time *after* the last request, so we're artificially making things take longer than they need to.  We probably want this at the beginning of the loop (which also will add a delay between the poll at line 28 and the first poll in the loop.)

ezimanyi: Also, is this deliberately quadratic backoff?  I'm less concerned about the particular backoff algorithm, but the max wait time here is < 1 min for 5 attempts...is that enough?

util/task_watcher.go
jtk54: Yep.

cmd/orca-tasks/task_watcher.go
jtk54: Agreed, I just removed all checks here and delegate to the normal while `for` loop.

cmd/orca-tasks/task_watcher.go
jtk54: `task == nil` handles the case where there is no retrieved task for any reason (network issues etc) which is why we're ignoring `resp` and `err` here.

cmd/orca-tasks/task_watcher.go
jtk54: I think the number of attempts is something we'll have to tune. This is just a heuristic that's been sufficient for now.

cmd/orca-tasks/task_watcher.go
ezimanyi: Makes sense :+1:

PR #122, opened 2019-02-08
fix(iap): Clear err variable after printing link.

duftler:
Without this fix, the error is still propagated even though the user has followed the displayed link.


jtk54:

PR #129, opened 2019-02-15
fix(pipeline/save): Fix setting pipeline type.

jtk54:


jtk54:

PR #130, opened 2019-02-19
fix(pipeline/exec): Output one pipeline execution after exec.

jtk54:


jtk54:

PR #133, opened 2019-02-22
feat(pipeline/execution): Adds list cmd for executions.

jtk54:


cmd/pipeline/execution/list.go
```diff
@@ -0,0 +1,118 @@
...
+		statuses = append(statuses, "CANCELED")
+	}
+	if len(statuses) > 0 {
+		query["statuses"] = strings.Join(statuses, ",")
+	}
+
+	if options.limit > 0 {
+		query["limit"] = options.limit
+	}
+	//fmt.Println(query)
```
ezimanyi: Unintentionally left here?

jtk54: Yep, removed.

jtk54:
We could add a new endpoint that handles the `succeeded` etc flags in the longer term but that is a bit of scope expansion. For now I'm going to drop a TODO to do the refactor/simplification of that endpoint.

jtk54:
Closes https://github.com/spinnaker/spinnaker/issues/3999.

PR #138, opened 2019-02-28
fix(pipeline-template): Accept STDIN in save cmd.

jtk54:
Also fails if id is not present.


jtk54:

PR #153, opened 2019-04-08
fix(gateclient): Removes spurious config warning.

jtk54:


jtk54:
It's not really an error, we are just defaulting the config. When folks are using spin for the first time, they see this warning each time even though the operations are succeeding, which is jarring.

jtk54:

PR #154, opened 2019-04-08
fix(app/save): Serializes providers as strings, not array.

jtk54:


jtk54:

PR #164, opened 2019-04-17
fix(gateclient): Converts reachable host check to http GET.

jtk54:
from TCP dial. Previous fix [here](https://github.com/spinnaker/spin/pull/152) was over-restrictive. @MrTrustor FYI.

PR #168, opened 2019-04-18
fix(gateapi): close body after http.Get to avoid leaking resources

sbwsg:
Per comment https://github.com/spinnaker/spin/pull/164#issuecomment-484340150

`gofmt` also tweaked a couple of lines here.

### DIFF
--- a/cmd/gateclient/client.go
+++ b/cmd/gateclient/client.go
@@ -22,6 +22,7 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"net/http/cookiejar"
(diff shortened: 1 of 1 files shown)


cmd/gateclient/client.go
```diff
@@ -400,6 +401,11 @@ func prompt() string {
 }
 
 func dialGate(gateClient *GatewayClient) error {
-	_, err := http.Get(gateClient.GateEndpoint())
-	return err
+	res, err := http.Get(gateClient.GateEndpoint())
+	if err != nil {
+		return err
+	}
+	io.Copy(ioutil.Discard, res.Body)
```
sbwsg: I was unclear initially why this drain procedure was suggested in the original comment. Reading around it seems like the go http lib may not reuse connections unless the body is drained completely before closing. This incurs a cost when working with https endpoints because the TLS handshake then takes time to re-establish for subsequent requests.

jtk54: :+1:

jtk54:

PR #174, opened 2019-05-09
add 'AllPipelinesForProjectUsingGET' to spin

RevinKey:


cmd/project/get.go
```diff
@@ -0,0 +1,94 @@
+// Copyright (c) 2018, Google, Inc.
```
jtk54: s/2018/2019

cmd/project/get.go
```diff
@@ -0,0 +1,94 @@
...
+		Short:   getProjectShort,
+		Long:    getProjectLong,
+		Example: getProjectExample,
+		RunE:    func(cmd *cobra.Command, args []string) error {
+			return getProject(cmd, options, args)
+		},
+	}
+
+	// Note that false here means defaults to false, and flips to true if the flag is present.
+	cmd.PersistentFlags().BoolVarP(&options.expand, "expand", "x", false, "expand app payload to include clusters")
```
jtk54: Remove if not used.

cmd/project/get.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+func NewGetCmd(prjOptions projectOptions) *cobra.Command {
+	options := GetOptions{
+		projectOptions: &prjOptions,
+		expand: false,
+	}
+
+	cmd := &cobra.Command{
+		Use:     "get",
```
jtk54: There's a specific endpoint to get a project [here](https://github.com/spinnaker/gate/blob/master/gate-web/src/main/groovy/com/netflix/spinnaker/gate/controllers/ProjectController.groovy#L39), so I think this needs reworked into something more appropriate with what the command is doing, e.g. `spin project get-pipelines`.

cmd/project/get.go
jtk54: This should also be your company name or author name instead of Google (since we didn't write it).

cmd/project/get.go
```diff
@@ -0,0 +1,94 @@
...
+			return fmt.Errorf("Encountered an error getting project, status code: %d\n", resp.StatusCode)
+		}
+	}
+
+	if err != nil {
+		return err
+	}
+	util.UI.JsonOutput(project, util.UI.OutputFormat)
+
+	// if options.expand {
```
jtk54: Cut commented code.

cmd/project/get.go
```diff
@@ -0,0 +1,94 @@
...
+		},
+	}
+
+	// Note that false here means defaults to false, and flips to true if the flag is present.
+	cmd.PersistentFlags().BoolVarP(&options.expand, "expand", "x", false, "expand app payload to include clusters")
+
+	return cmd
+}
+
+func getProject(cmd *cobra.Command, options GetOptions, args []string) error {
```
jtk54: Should have a unit test.

PR #176, opened 2019-05-14
fix(client): Reverts Gate liveness check.

jtk54:
The Gate liveness check needs rethought since it doesn't handle authentication cases and is breaking for forwarded ports. Reverting for now.


jtk54:

PR #177, opened 2019-05-14
fix(client): Reverts Gate liveness check.

jtk54:
@duftler FYI.


jtk54:
Whoops, against master not release-1.13.x

Issue #184, opened 2019-05-21
Saving template when gate is down produces non friendly error message


eh-am:
(With gate down)
Running `spin pipeline-templates save --file my-template.json` gives

```
panic: runtime error: invalid memory address or nil pointer dereference            
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x872778]
                      
goroutine 1 [running]:                                                                                                                          
github.com/spinnaker/spin/cmd/pipeline-template.savePipelineTemplate(0xc000174f00, 0xd66de8, 0x0, 0x0, 0x7ffe87725677, 0x20, 0x0, 0x0, 0x0, 0x0)      
        /var/lib/jenkins/workspace/Build_PrimaryArtifacts/build_input/build_spin/src/github.com/spinnaker/spin/cmd/pipeline-template/save.go:94 +0x258
github.com/spinnaker/spin/cmd/pipeline-template.NewSaveCmd.func1(0xc000174f00, 0xc0000c2a80, 0x0, 0x2, 0x0, 0x0)                                     
        /var/lib/jenkins/workspace/Build_PrimaryArtifacts/build_input/build_spin/src/github.com/spinnaker/spin/cmd/pipeline-template/save.go:47 +0x52
github.com/spf13/cobra.(*Command).execute(0xc000174f00, 0xc0000c29e0, 0x2, 0x2, 0xc000174f00, 0xc0000c29e0)                      
        /var/lib/jenkins/workspace/Build_PrimaryArtifacts/build_input/build_spin/src/github.com/spf13/cobra/command.go:762 +0x473
github.com/spf13/cobra.(*Command).ExecuteC(0xc0000f7b80, 0xc00016cc80, 0xc00016c000, 0xc0000d6600)                               
        /var/lib/jenkins/workspace/Build_PrimaryArtifacts/build_input/build_spin/src/github.com/spf13/cobra/command.go:850 +0x2f8
github.com/spf13/cobra.(*Command).Execute(0xc0000f7b80, 0xc0000b6008, 0xc0000f7b80)                                             
        /var/lib/jenkins/workspace/Build_PrimaryArtifacts/build_input/build_spin/src/github.com/spf13/cobra/command.go:800 +0x2b
github.com/spinnaker/spin/cmd.Execute(0xa03d00, 0xc0000b6008, 0xc0000a3f88, 0x405620)                                              
[...]

PR #185, opened 2019-05-21
feat(client): Checks for Gate liveness before executing cmd.

jtk54:
@MrTrustor PTAL and test run as well please.


jtk54:
Fixes https://github.com/spinnaker/spin/issues/184.

PR #191, opened 2019-06-14
feat(docker): Add Dockerfile for building Docker images

brantburnett:
This allows a complete build using Docker, or building an image from an already compiled copy using using Dockerfile.slim.

This partially addresses https://github.com/spinnaker/spinnaker/issues/4531. Further modifications are probably required to https://github.com/spinnaker/spinnaker/blob/master/dev/buildtool/spin_commands.py to address release publication.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,19 @@
+FROM golang:1.12 as build
+ARG VERSION
+
+WORKDIR /app
+COPY ./ ./
+
+RUN ./build.sh --go-os linux --go-arch amd64 --version "${VERSION:-}"
+
+
+
+FROM alpine
+
+RUN apk update \
+    && apk add ca-certificates \
+    && rm -rf /var/cache/apk/*
+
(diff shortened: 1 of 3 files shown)


brantburnett:
@louisjimenez Agreed, but I didn't want to leave release.sh in a bad state with my changes. If it's really deprecated, perhaps we should remove it from the repo instead?

louisjimenez:
Definitely, I added this issue to clean things up, https://github.com/spinnaker/spin/issues/195

brantburnett:
@louisjimenez I've reverted release.sh for now

louisjimenez:
Thanks @brantburnett . Would getting what remains of this PR merged be useful for your workflow? I believe changes would still be needed to the `spin_commands` to get the image built and published but let me know if I am missing a use case. Thanks!

brantburnett:
@louisjimenez Yes please, this is step #1

PR #217, opened 2019-09-17
fix(cmd): Remove flaky checks for successful executions

gregjones:
Checking for exactly one RUNNING execution isn't reliable for a few
of reasons:
- for a pipeline that allows multiple simultaneous executions, there can
  be more than one running at a time
- for a pipeline that doesn't allow multiple executions, there is no
  check that the returned execution is the one that was just triggered,
  instead of one that was executing previously
- if a pipeline starts and finishes in between the retries, while the
  check is sleeping, then it will report failure despite having succeeded

### DIFF
--- a/cmd/pipeline/execute.go
+++ b/cmd/pipeline/execute.go
@@ -18,7 +18,6 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
-	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/spinnaker/spin/cmd/gateclient"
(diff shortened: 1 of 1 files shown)


xavileon:
Thanks!
@spinnaker/reviewers PTAL

cmd/pipeline/execute.go
```diff
@@ -107,34 +106,5 @@ func executePipeline(cmd *cobra.Command, options ExecuteOptions) error {
...
-	}
-	if len(executions) == 0 {
-		return fmt.Errorf("Unable to start any executions, server response was: %v", resp)
-	}
-
-	if len(executions) > 1 {
-		return fmt.Errorf("Started more than one execution: %v", executions)
-	}
-
-	util.UI.JsonOutput(executions[0], util.UI.OutputFormat)
```
louisjimenez: Could we add something to indicate to the user that the request succeeded, similar to https://github.com/spinnaker/spin/blob/master/cmd/pipeline/save.go#L117

gregjones: Ah, yes, that makes sense: https://github.com/spinnaker/spin/pull/217/files#diff-217e0cd0e364fb48908d6a7e8e520cf6R110

(Thanks for the quick review!)

PR #218, opened 2019-09-18
chore(gateapi): Update gate api for 1.16.1

louisjimenez:
Closes #4927. This is 99% generated code from Swagger. This is a fairly large diff since the last time the gate client api was updated was for Spinnaker 1.13 and since than the version of Swagger used by kork has also changed.

I tested out these changes by running the test suite and doing some smoke testing.


cmd/canary/canary-config/save.go
```diff
@@ -78,7 +78,7 @@ func saveCanaryConfig(cmd *cobra.Command, options SaveOptions) error {
 	var saveErr error
 	if resp.StatusCode == http.StatusOK {
 		_, saveResp, saveErr = gateClient.V2CanaryConfigControllerApi.UpdateCanaryConfigUsingPUT(
-			gateClient.Context, templateId, templateJson, map[string]interface{}{})
+			gateClient.Context, templateJson, templateId, map[string]interface{}{})
```
louisjimenez: This was the one non automated change. The order of the arguments to the api helper function had to be swapped.

plumpy: Just curious, but why was that? Did the order change in the gate API or something?

louisjimenez: Created an issue to track this, https://github.com/spinnaker/spinnaker/issues/4931

student-06:
Hi, @louisjimenez @plumpy 
`spin pipeline execute --name execute-check --application test4 --output=jsonpath='{.id}'`
This is not working in `1.12` version but used to work in `1.7.2`. Is there anything I missing or is this an issue ?

louisjimenez:
@student-06 is the pipeline executing? The output of this command did change in 1.12, https://github.com/spinnaker/spin/pull/217.

PR #227, opened 2019-10-23
feat(canary): Adds retrospective cmd for local config testing.

jtk54:


jtk54:
PTAL @louisjimenez @duftler I think they finally revoked my 'add reviewers' privileges...

fieldju:
Am I missing something or is there not a place to set extended scope parameters?

fieldju:
I am also not seeing where I would select which metrics/storage account I want to use, which in my case where I have multiple accounts is required else it defaults to the first one.

jtk54:
Fair enough on the optional params, adding support and will update PR.

jtk54:
Updated, PTAL.

cmd/canary/canary-config/retro.go
```diff
@@ -0,0 +1,236 @@
...
+	*canaryConfigOptions
+	output                  string
+	canaryConfigFile        string
+	controlGroup            string
+	controlLocation         string
+	experimentGroup         string
+	experimentLocation      string
+	startInstant            string
+	endInstant              string
+	extendedScopeParamsFile string
```
fieldju: What do you think about the idea of using [repeat flags](https://github.com/spf13/cobra/issues/661#issuecomment-377684634) for this instead of a file?

eg `spin retrocmd -ep '_scope_key=autoscaling-group-name' -ep '_location_key=region'`

then that that generates

`[]string{"_scope_key=autoscaling-group-name", "_location_key=region"}`

Which can be used to generate the JSON map for extended scope params.

jtk54: `pflag` has a nice `StringToStringVar` flag that lets you specify a map explicitly with `--my-flag="key1=value1,key2=value2,..."` so I went with that.

PR #232, opened 2019-11-13
feat(auth): Allow command line credentials when using ldap authN

titirigaiulian:
Allowing command line params for login simplifies integrations of spin cli in scripts or other wrappers for automation purposes.


titirigaiulian:
It should work adding env vars in the config file. Closing this pr

PR #234, opened 2019-11-19
feat(cmd): add `account` sub command for querying configured accounts

jwelch92:
The `account` subcommand has `get` and `list` commands. Useful for getting all available deployment targets when generating pipelines.

### DIFF
--- a/cmd/account/get.go
+++ b/cmd/account/get.go
@@ -1,0 +1,82 @@
+// Copyright 2019 New Relic Corporation. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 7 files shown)


jwelch92:
Added a fix for the builds not working due to go 1.11 not auto enabling go modules. I can split that fix into a separate PR if needed.

cmd/account/get.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/spinnaker/spin/util"
+	"net/http"
+
+	"github.com/spf13/cobra"
+	"github.com/spinnaker/spin/cmd/gateclient"
+)
+
+type GetOptions struct {
+	*accountOptions
+	expand bool
```
louisjimenez: I believe expand can be omitted for getting a single account

jwelch92:
@louisjimenez Updated. Removed `expanded` from get, sorted imports across changed files.

jwelch92:
Thanks @louisjimenez !!

PR #241, opened 2020-02-27
fix(jsonpath): Return all matches from jsonpath, not just first

sikevux:
Currently passing something like `"jsonpath={[*]..id},{[*]..application}"` to
--output only returns the id. This change returns all the matches instead of
just the first one.

Example:
Given the pipeline list json in pipelineListJson, we would currently return
`"pipeline1"` if the user passed `"jsonpath={[*]..id},{[*]..application}"` to --output

With this change we would print `"pipeline1,app"` instead, as that's how kubectl
would print it.

When you have multiple pipelines the output would be:
`"pipeline1 pipeline2 pipeline3,app app app"`

### DIFF
--- a/util/cli_ui.go
+++ b/util/cli_ui.go
@@ -103,12 +103,8 @@ func (u *ColorizeUi) JsonOutput(input interface{}, outputFormat *output.OutputFo
 			u.Error(fmt.Sprintf("%v", err))
 		}
 
-		prettyStr, err := json.MarshalIndent(jsonValue, "", " ")
-		if err != nil {
-			u.Error(fmt.Sprintf("%v", err))
-		}
 		// unquote since go quotes the string if the bytes is a string.
-		u.Output(u.colorize(u.unquote(string(prettyStr)), u.OutputColor))
+		u.Output(u.colorize(u.unquote(jsonValue.String()), u.OutputColor))
 	}
 }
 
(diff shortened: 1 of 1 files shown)


util/cli_ui.go
```diff
@@ -131,8 +131,14 @@ func (u *ColorizeUi) parseJsonPath(input interface{}, template string) (interfac
...
-	if values != nil && len(values) > 0 && len(values[0]) > 0 {
-		return values[0][0].Interface(), nil
+	if values != nil && len(values) > 0 {
+		valueStrings := []string{}
+		for idx := range values {
+			for val := range values[idx] {
+				valueStrings = append(valueStrings, fmt.Sprintf("%v", values[idx][val].Interface()))
+			}
+		}
+		return strings.Join(valueStrings, ""), nil
```
louisjimenez: Given the example expression you mentioned `"jsonpath={[]..id},{[]..application}"`, what is the expected output  for an application with multiple pipelines?

sikevux: With `jsonpath={[*]..id},{[*]..application}` it would be something like `uuid1uuid2uuid3,name1name2name3`.

Without this change the output would be:
`uuid1`

louisjimenez: Thoughts on adding a separator? Kubectl seems to use a space between values.

sikevux: Yeah, I'm getting more partial to their way of doing it the more I use this patch. I'll make some changes

karlkfi:
This change is very reverse incompatible (not just a fix), and not in a good way, IMO. 

Ex: 
```
spin -k app get "my-app" --output "jsonpath={.permissions}"
```

now prints
```
map[EXECUTE:[admin-group] READ:[admin-group user-group] WRITE:[admin-group]]
```

instead of
```
{
 "EXECUTE": [
  "admin-group"
 ],
 "READ": [
  "admin-group",
  "user-group"
 ],
 "WRITE": [
  "admin-group"
 ]
}
```

sikevux:
> This change is very reverse incompatible (not just a fix), and not in a good way, IMO.
> 
> Ex:
> 
> ```
> spin -k app get "my-app" --output "jsonpath={.permissions}"
> ```
> 
> now prints
> 
> ```
> map[EXECUTE:[admin-group] READ:[admin-group user-group] WRITE:[admin-group]]
> ```
> 
> instead of
> 
> ```
> {
>  "EXECUTE": [
>   "admin-group"
>  ],
>  "READ": [
>   "admin-group",
>   "user-group"
>  ],
>  "WRITE": [
>   "admin-group"
>  ]
> }
> ```

Sorry for the breakage. The intention was not to make incompatible changes but rather to fix what you get wen you run jsonpath queries like this:
```
spin app list --output 'jsonpath={[*]..name}'
```

In earlier versions you would only get one result, even tho you had multiple applications. That's not jsonpath compatible either.

The fastest solution is probably to back out this change and then create a new fix for the jsonpath incompatibility without adding unintended incompatibility in other places.

PR #242, opened 2020-03-23
chore(build): Rename workflow

jonsie:


.github/workflows/ci.yml
```diff
@@ -1,4 +1,4 @@
-name: Spin CI
+name: Branch Build
```
jonsie: This is to support some build status doc changes.

PR #243, opened 2020-03-24
Add YAML Support

karlkfi:
JSON may be what Spinnaker's API uses, but it's tedious to write manually and harder for humans to read and manage. So I added YAML support in the CLI to covert back and forth easily. This way we can store our Spinnaker pipeline configs in YAML and do GitOps, the same way we do for Kubernetes manifests.

- Overload --output command to support --output yaml
- Replace json parser with yaml parser (json is a subset of yaml)
- Use Kubernetes' yaml parser, which converts through JSON to support annotations
- Enable strict parsing mode (dissallow duplicate map keys. better error messages)
- Wrap some error messages with more context

Tests were broken after I enabled strict parsing, so I had to modify the pipeline-template plan tests so stop using a duplicate map key to test an invalid config. I changed it to an unknown key instead. After this, I all the tests passed, using the yaml parser.

I also did some cleanup of the codebase while I was in there (go fmt, goimports, test command refactor). If it makes review harder, you can look at each commit separately.

While manually testing, I realized that. `--output yaml` was tedious to type, so i added `-o` as a shortcut. While doing do, I also realized the app save subcommand was missing some common shorts, so I added `-a` for `--application-name` and `-f` for `--file`.

### DIFF
--- a/cmd/account/get_test.go
+++ b/cmd/account/get_test.go
@@ -16,12 +16,13 @@ package account
 
 import (
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
 	"strings"
 	"testing"
 
-	"github.com/spinnaker/spin/util"
+	"github.com/spinnaker/spin/cmd"
 )
 
 // GateServerFail spins up a local http server that we will configure the GateClient
@@ -40,15 +41,12 @@ const (
 func TestAccountGet_basic(t *testing.T) {
 	ts := testGateAccountGetSuccess()
 	defer ts.Close()
-	currentCmd := NewGetCmd(accountOptions{})
-	rootCmd := util.NewRootCmdForTest()
-	accCmd := NewAccountCmd()
-	accCmd.AddCommand(currentCmd)
-	rootCmd.AddCommand(accCmd)
+	rootCmd, options := cmd.NewCmdRoot(ioutil.Discard, ioutil.Discard)
+	rootCmd.AddCommand(NewAccountCmd(options))
 
 	args := []string{"account", "get", ACCOUNT, "--gate-endpoint=" + ts.URL}
 	rootCmd.SetArgs(args)
-	_, err := util.ExecCmdForTest(rootCmd)
+	err := rootCmd.Execute()
 	if err != nil {
 		t.Fatalf("Command failed with: %s", err)
 	}
@@ -57,14 +55,11 @@ func TestAccountGet_basic(t *testing.T) {
 func TestAccountGet_flags(t *testing.T) {
 	ts := testGateAccountGetSuccess()
 	defer ts.Close()
-	currentCmd := NewGetCmd(accountOptions{})
-	rootCmd := util.NewRootCmdForTest()
-	accCmd := NewAccountCmd()
-	accCmd.AddCommand(currentCmd)
-	rootCmd.AddCommand(accCmd)
+	rootCmd, options := cmd.NewCmdRoot(ioutil.Discard, ioutil.Discard)
+	rootCmd.AddCommand(NewAccountCmd(options))
 	args := []string{"account", "get", "--gate-endpoint", ts.URL} // Missing positional arg.
 	rootCmd.SetArgs(args)
-	_, err := util.ExecCmdForTest(rootCmd)
+	err := rootCmd.Execute()
 	if err == nil { // Success is actually failure here, flags are malformed.
 		t.Fatalf("Command failed with: %s", err)
 	}
@@ -74,15 +69,12 @@ func TestAccountGet_malformed(t *testing.T) {
 	ts := testGateAccountGetMalformed()
 	defer ts.Close()
 
-	currentCmd := NewGetCmd(accountOptions{})
-	rootCmd := util.NewRootCmdForTest()
-	accCmd := NewAccountCmd()
-	accCmd.AddCommand(currentCmd)
-	rootCmd.AddCommand(accCmd)
+	rootCmd, options := cmd.NewCmdRoot(ioutil.Discard, ioutil.Discard)
+	rootCmd.AddCommand(NewAccountCmd(options))
 
 	args := []string{"account", "get", ACCOUNT, "--gate-endpoint=" + ts.URL}
 	rootCmd.SetArgs(args)
-	_, err := util.ExecCmdForTest(rootCmd)
+	err := rootCmd.Execute()
 	if err == nil { // Success is actually failure here, return payload is malformed.
 		t.Fatalf("Command failed with: %d", err)
 	}
@@ -92,15 +84,12 @@ func TestAccountGet_fail(t *testing.T) {
 	ts := GateServerFail()
 	defer ts.Close()
 
-	currentCmd := NewGetCmd(accountOptions{})
-	rootCmd := util.NewRootCmdForTest()
-	accCmd := NewAccountCmd()
-	accCmd.AddCommand(currentCmd)
-	rootCmd.AddCommand(accCmd)
+	rootCmd, options := cmd.NewCmdRoot(ioutil.Discard, ioutil.Discard)
+	rootCmd.AddCommand(NewAccountCmd(options))
 
 	args := []string{"account", "get", ACCOUNT, "--gate-endpoint=" + ts.URL}
 	rootCmd.SetArgs(args)
-	_, err := util.ExecCmdForTest(rootCmd)
+	err := rootCmd.Execute()
 	if err == nil { // Success is actually failure here, return payload is malformed.
 		t.Fatalf("Command failed with: %d", err)
 	}
--- a/cmd/account/get_test.go
+++ b/cmd/account/get_test.go
@@ -15,6 +15,7 @@
 package account
 
 import (
+	"bytes"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -22,12 +23,14 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/andreyvit/diff"
 	"github.com/spinnaker/spin/cmd"
+	"github.com/spinnaker/spin/util"
 )
 
-// GateServerFail spins up a local http server that we will configure the GateClient
+// testGateFail spins up a local http server that we will configure the GateClient
 // to direct requests to. Responds with a 500 InternalServerError.
-func GateServerFail() *httptest.Server {
+func testGateFail() *httptest.Server {
 	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		// TODO(jacobkiefer): Mock more robust errors once implemented upstream.
 		http.Error(w, "Internal Server Error", http.StatusInternalServerError)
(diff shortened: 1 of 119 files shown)


karlkfi:
Updated commit messages to match 'conventional' format.

karlkfi:
Clarified commit summaries for changelog use.

karlkfi:
Refactored all the rest of the tests to buffer cmd output. None of them are currently using the output, so we could hypothetically pipe to /dev/null, but my hope is that this allows tests to validate the output in the future.

plumpy:
Your last commit has
> Also ran go fmt & goimports to clean up the code

Could you do that in a separate commit? It introduces a _huge_ amount of changes, so makes it very difficult to see what the _real_ changes you made were...

karlkfi:
@plumpy done.

cmd/output/output.go
```diff
@@ -24,13 +24,17 @@ import (
 type OutputFormat struct {
 	// JsonPath specifies a subpath of the output to extract data from
 	JsonPath string
+	Yaml     bool
 }
```
plumpy: Instead of just saving these flags, could you instead do something like this

```
type OutputConverter interface {
  convert(input interface{}) (string, error)
}
```

And then put the three different `OutputFormat` implementations in this file.

cmd/pipeline-template/plan_test.go
```diff
@@ -223,11 +223,11 @@ const testPlanConfig = `
             "enabled": true,
             "pubsubSystem": "google",
             "subscription": "super-derp",
-            "subscription": "super-derp",
             "subscriptionName": "super-derp",
             "source": "jack",
             "attributeConstraints": {},
-            "payloadConstraints": {}
+            "payloadConstraints": {},
+						"invalid-key": "whatever"
```
plumpy: Curious about the reason for this change? This implies that perhaps the refactoring causes some subtle change in input parsing behavior... can you explain what that change is?

karlkfi: > Tests were broken after I enabled **strict parsing**, so I had to modify the pipeline-template plan tests so stop using a duplicate map key to test an invalid config. I changed it to an unknown key instead. After this, I all the tests passed, using the yaml parser.

cmd/output/output.go
karlkfi: I'm not sure I follow. Why do we need multiple?

cmd/pipeline-template/plan_test.go
plumpy: So enabling strict parsing could mean that some pipelines that previously worked fine for users will no longer work, correct? I suspect it's not a huge deal, but we should probably add something to the release notes if that's the case...

cmd/output/output.go
plumpy: Sorry, I meant three different `OutputConverters`... one for YAML, one for plain JSON, and one for JSON with a `JsonPath`. (Or maybe those last two can be the same?) i.e. so change this:
```
func ParseOutputFormat(outputFormat string) (*OutputFormat, error)
```
to
```
func CreateOutputConverter(outputFormat string) (*OutputConverter, error)
```

cmd/pipeline-template/plan_test.go
karlkfi: The API should still error on save if the input is invalid. Ideally, this just moves the error from the server side to the client side.

That said, the Spinnaker API is pretty bad at validating content, IME. So it's possible this causes some hard errors that were previously being ignored.

karlkfi:
> Would love to see at least one test that shows YAML input working and another showing YAML output...

Turns out I can't write a test that validates output because I need to first rewrite the UI to not be global. Right now the cobra.Command and the util.UI have different output writers. So while I had buffered the Command output, I hadn't buffered the UI output. 

Buffering the UI output is turning into a giant refactor because:
- UI is global and thus can't be changed in a thread-safe way (test can run in parallel, hypothetically)
- The UI is being updated AFTER the cmd.Execute() by the GateClient, which is doing flag parsing and InitUI. (This is a terrible place to be parsing global flags... someone apparently just shoehorned it in here because the GatClient flags were being parsed here.)
- Injecting the UI into every command will require depending on context from a parent command. But right now the package dependencies all face in the wrong way: parent packages depending on child packages. So in order to inject the UI into the subcommands, I have to first reverse the package dependencies.
- Reversing the package dependencies is itself a huge refactor, because right now the parent commands add all the child commands. So I need to pull out some sort of command assembler that depends on all the cmd packages without being their parent.

And THEN I need to write some tests, because apparently none of the output is being tested right now, which means all those refactors are pretty unsafe. 

I'm not sure I want to backfill output tests for every command output, but I can kick it off, if I survive the above refactoring...

With this much refactoring, and lack of tests, it's probably gonna need a major revision bump just out of caution, because we wont actually know if any output broke.

karlkfi:
I pushed a very large refactor to remove the global UI and allow buffering of all output.

New YAML tests still TODO.

karlkfi:
Alright, I've added a bunch of tests for both json and yaml that validate command output and request body sent to the gate server.

This was a lot of work. I hope the PR isn't too large to get approved. With this much refactoring, it's going to be a serious pain to have to rebase if other PRs are merged first.

@plumpy This is ready for another review pass.

cmd/output/output.go
karlkfi: I refactored this to be 3 different implementations of `type OutputFormater func(interface{}) ([]byte, error)` instead of a single implementation with if statements.

cmd/output/output.go
```diff
@@ -15,31 +15,98 @@
...
+
+func MarshalToYaml(input interface{}) ([]byte, error) {
+	pretty, err := yaml.Marshal(input)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to marshal to yaml: %v", err)
+	}
+	return pretty, nil
+}
+
+// TODO: is this nessesary when using bytes.Bytes() instead of bytes.String() ?
```
plumpy, line 92: Looks like you can maybe remove this, since you are using `bytes.Bytes()`?

cmd/output/output.go
```diff
@@ -15,31 +15,98 @@
...
+// (https://kubernetes.io/docs/reference/kubectl/jsonpath/).
+func MarshalToJsonPathWrapper(expression string) OutputFormater {
+	expr := expression
+	// aka MarshalToJsonPath
+	return func(input interface{}) ([]byte, error) {
+		jp := jsonpath.New("json-path")
+		buffer := new(bytes.Buffer)
+
+		if err := jp.Parse(expr); err != nil {
+			return buffer.Bytes(), fmt.Errorf("Failed to parse jsonpath expression: %v", err)
```
plumpy, line 71: should you return `nil` instead of `buffer.Bytes()`, especially since the buffer is guaranteed to be empty here anyway? I suppose it might possibly have some content in the next return statement, but it still feels weird to return unfinished content _and_ an error... or is that normal in Go?

cmd/output/output.go
```diff
@@ -15,31 +15,98 @@
...
+// TODO: is this nessesary when using bytes.Bytes() instead of bytes.String() ?
+func unquote(input []byte) []byte {
+	input = bytes.TrimLeft(input, "\"")
+	input = bytes.TrimRight(input, "\"")
+	return input
+}
+
+// parseJsonPath finds the values specified in the input data as specified with the template.
+// This leverages the kubernetes jsonpath libs (https://kubernetes.io/docs/reference/kubectl/jsonpath/).
+func parseJsonPath(input interface{}, template string) (*bytes.Buffer, error) {
```
plumpy, line 101: looks like this is unused... probably got refactored into the closure inside `MarshalToJsonPathWrapper`?

cmd/account/get_test.go
```diff
@@ -146,14 +160,26 @@ const malformedAccountGetJson = `
 }
 `
 
+//TODO(karlkfi): Why is the json consistently ordered but not alphabetical?
```
plumpy, line 163: Generally I'd write a test like this by parsing the output JSON as then asserting properties about it, e.g. (pseudocode) `assert json.get("name") == "account"` for all 7 properties, and then `assert json.properties().count() == 7`. That insulates you from changes in the exact formatting of the output, i.e. if you upgrade a library and the sorting changes or it starts adding quotes around the YAML values or something...

But that would be a massive change and I'm not trying to imply you should do that, just FYI...

plumpy:
Just LMK when you're done with the few small things I mentioned and then I'll go ahead and merge it. Thank you!!

cmd/output/output.go
plumpy: (er... to clarify, I meant remove the TODO since unfortunately it looks like you will need to keep using this `unquote` function)

karlkfi: Seems like https://github.com/spinnaker/spin/pull/241 broke jsonpath in a non-reverse-compatible way (with a "fix" commit). Instead of fixing that whole mess, I'm gonna remove my TODO and leave it doing the unquote.

This also means I can't add all the jsonpath tests I wanted, because I don't actually know what the desired behavior is now.

cmd/application/delete.go
```diff
@@ -80,19 +80,19 @@ func deleteApplication(cmd *cobra.Command, args []string) error {
...
 	}
 
-	err = orca_tasks.WaitForSuccessfulTask(gateClient, taskRef, 5)
+	err = orca_tasks.WaitForSuccessfulTask(options.GateClient, taskRef, 5)
 	if err != nil {
 		return err
 	}
 
-	util.UI.Output(util.Colorize().Color(fmt.Sprintf("[reset][bold][green]Application deleted")))
+	options.Ui.Output("Application deleted")
```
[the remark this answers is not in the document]
karlkfi: Yes. will fix

cmd/output/output.go
karlkfi: sure. it probably doesn't matter but nil might be clearer.

cmd/output/output.go
karlkfi: yup. will delete.

cmd/account/get_test.go
karlkfi: There's no real ordering validation happening elsewhere, due to the general lack of UI and output tests... So I was trying to make sure there was at least ordering test coverage somewhere...

I have no idea why the API client mutates the order received from the server tho... when I make this alphabetical as what `testGateAccountGetSuccess()` returns, the test fails because the return from the `spin account get` command is sorted in some consistent, non-alphabetical way. I'm pretty sure this is a client bug, but I haven't found it...

karlkfi: So I found the weird consistent ordering. It's in the swagger API definition:
https://github.com/spinnaker/spin/blob/master/gateapi/api/swagger.yaml#L6055-L6088

And it gets propagated to the AccountDetails go object:
https://github.com/spinnaker/spin/blob/master/gateapi/account_details.go#L12

And the client returns an AccountDetails object instead of a map[string]interface{}:
https://github.com/spinnaker/spin/blob/master/gateapi/credentials_controller_api.go#L48

So the JsonOutput formats it in the order from the Object... crazy.

karlkfi:
Two remaining issues, pulling out of the inline comments:

1. #241 broke jsonpath in a reverse incompatible way. I can't add tests for non-primitive jsonpath results, because they aren't printed out in json format anymore, and I'm not sure that's intentional. The commit that changed it was called a fix but it's definitely more of a breaking change and feature.
2. Output field sorting in inconsistent across commands, but no one knew because there weren't any previous tests validating them. I added some tests, but they may make it look like the order is intentional. I left some comments about it in the code, but I think there should really be a follow-up PR to fix the ordering to be consistently alphabetical, instead of ordering the way the swagger.yaml is ordered (which seems likely unintentional).

plumpy:
Merging this now. Thanks so much!! :tada:

karlkfi:
@plumpy looks like squash merging lost the commit message prefixes (seems a common problem, looking at master). Does that mean all PR titles need the same prefixes, like `feat(cmd)`?

Does it need to be fixed on master? Force push is bad for forks :(

plumpy:
I think Spinnakerbot complains if your sub-commits don't have the proper prefixes, but the only one that _really_ matters is the main commit that goes onto master, i.e. the squash message.

In the end, it's probably not a huge deal. Those messages are used to generate the version numbers (i.e. so a series of `fix` commits, will only bump the patch version, not the minor version) but I don't think it'll cause huge problems.

PR #244, opened 2020-04-03
refactor base gate http client to be usable outside spin cli

coryb:
Hey all, I am working on a [spinnaker managed delivery tool](https://github.com/Spinnaker/md-lib-go) where I was hoping to reuse the ~/.spin/config file and authentication code that is in this project.  This PR tries to untangle the http client authentication code from the cli/flags setup so that it can be used independently.

I don't believe there are any functional changes for the `spin` project, just moving some of the code around and creating a few more exported functions.

This is currently how I am using this code:
https://github.com/spinnaker/md-lib-go/blob/master/cmd/spinmd/main.go#L41

I am open to any suggestions.  Thanks!

### DIFF
--- a/cmd/gateclient/client.go
+++ b/cmd/gateclient/client.go
@@ -19,9 +19,10 @@ import (
 	"bufio"
 	"context"
 	"crypto/rand"
+	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
-	"errors"
+	"encoding/base64"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -34,21 +35,18 @@ import (
 	"strings"
 	"syscall"
 
+	"github.com/mitchellh/go-homedir"
+	"github.com/pkg/errors"
 	"github.com/spinnaker/spin/cmd/output"
 	"github.com/spinnaker/spin/config"
+	"github.com/spinnaker/spin/config/auth"
 	iap "github.com/spinnaker/spin/config/auth/iap"
-	"github.com/spinnaker/spin/version"
-
-	"github.com/mitchellh/go-homedir"
-	"sigs.k8s.io/yaml"
-
-	"crypto/sha256"
-	"encoding/base64"
-
 	gate "github.com/spinnaker/spin/gateapi"
+	"github.com/spinnaker/spin/version"
 	"golang.org/x/crypto/ssh/terminal"
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
+	"sigs.k8s.io/yaml"
 )
 
 const (
@@ -103,6 +101,7 @@ func NewGateClient(ui output.Ui, gateEndpoint, defaultHeaders, configLocation st
 		gateEndpoint:     gateEndpoint,
 		ignoreCertErrors: ignoreCertErrors,
 		ui:               ui,
+		Context:          context.Background(),
 	}
 
 	err := userConfig(gateClient, configLocation)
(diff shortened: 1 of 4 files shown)


util/cli_ui.go
```diff
@@ -39,12 +39,18 @@ type ColorizeUi struct {
...
 }
 
-var UI *ColorizeUi
+var UI = ColorizeUi{
+	Colorize: Colorize(),
+	Ui: &cli.BasicUi{
+		Writer:      os.Stdout,
+		ErrorWriter: os.Stderr,
+	},
+}
```
coryb, line 48: This change was made so we don't get a panic for de-referencing a nil pointer if a project has not called InitUI for message like:
```
util.UI.Error("Could not initialize http client, failing.")
```

config/auth/authconfig.go
```diff
@@ -25,12 +25,13 @@ import (
...
 // Config is the CLI's authentication configuration.
 type Config struct {
-	Enabled bool           `yaml:"enabled"`
-	X509    *x509.Config   `yaml:"x509,omitempty"`
-	OAuth2  *oauth2.Config `yaml:"oauth2,omitempty"`
-	Basic   *basic.Config  `yaml:"basic,omitempty"`
-	Iap     *config.Config `yaml:"iap,omitempty"`
-	Ldap    *ldap.Config   `yaml:"ldap,omitempty"`
+	Enabled          bool           `yaml:"enabled"`
+	IgnoreCertErrors bool           `yaml:"ignoreCertErrors"`
```
coryb, line 29: Added `IgnoreCertErrors` config option to mimic the `ignoreCertErrors` flag via configuration. My company's Spinnaker install uses mtls with internal certificate authority, so we have to set this for all internal requests.

coryb:
Hey all, trying to figure out if there is a moral objection to the abstraction of your http client logic for reuse in other spinnaker related projects?  If so I can copy/paste and move along.  If you are open to the idea I can spend the time to rebase and update the PR to account for recent changes.  Thanks.

coryb:
Another idea I had that might be more or less appealing is to factor the http client + auth logic out into a standalone library that both your spin cli and my tool could consume?  So not sure if that would be a more interesting direction.

luispollo:
> I don't believe there are any functional changes for the `spin` project, just moving some of the code around and creating a few more exported functions.

Have you tried building and running `spin` locally for a quick sanity check? I'm not familiar with the automated test coverage for the CLI.

cmd/gateclient/client.go
```diff
@@ -96,6 +97,70 @@ func (m *GatewayClient) GateEndpoint() string {
 	return m.Config.Gate.Endpoint
 }
 
+func NewHTTPClient(ctx context.Context, cfg *config.Config) (*http.Client, context.Context, error) {
```
ezimanyi, line 100: The pattern of returning a context from `NewHTTPClient` seems strange to me...generally I'd expect this to accept a context, but not to return one.

ezimanyi:
Overall I think this is reasonable and makes the code hear cleaner as well.  That being said, I do want to caution that `spin-cli` doesn't really have a guarantee of compatibility for library consumers as it's primarily intended to support building the binary.  Maybe at some point we could think of that...but at this point I think anyone consuming this as a library should probably expect breaking changes and be fine dealing with those on upgrades.

So while I do think this is better, and would be glad to merge this to make it easier for your project to talk to Spinnaker, I do want to reasonably set expectations about the stability of our internal APIs.

cmd/gateclient/client.go
coryb: Yeah, for sure, not ideal. The problem is that the original code uses context state values, which returns context copy.  I was unsure how else to accomplish this without returning the modified context.  Like here:
https://github.com/spinnaker/spin/pull/244/files#diff-146916aad3615e73605759c9ea128128R357

I will think about alternate strategies, not sure we have many other than to capture into some client wrapper struct like the current code does.

coryb:
Thanks for the look @ezimanyi, I understand what you are saying. I wonder if it would make more sense to just factor out the common http client + auth logic into a stand-alone library that both our tools could consume, that way forward might be less brittle long term.

cmd/gateclient/client.go
coryb: I separated out the code that modified context state into a single `ContextWithAuth` so now it is more clear without the ctx coping everywhere.

coryb:
Hey @ezimanyi, I have updated the code to try to handle the recent changes.  I have updated the managed delivery project to use this new refactor, it seems pretty reasonable to me:
https://github.com/spinnaker/md-lib-go/blob/master/cmd/spinmd/main.go#L41-L58

There were a few hoops to jump through as I didnt think it made sense to add a `ui` object to these apis and for the errors that were previous logged with `ui.Error` I wrap them with `errors.Wrap` instead so the spin-cli can unwrap them and display them as before.

Let know if you have any more questions or concerns.  Thanks for your time!

PR #245, opened 2020-04-07
fix(application): Add shorthand f for file in application save

sikevux:
This makes application save match pipeline and pipeline-template


plumpy:
Thanks, this is actually already included as part of #243, which should get merged soon...

PR #246, opened 2020-04-30
fix(jsonpath): revert jsonpath change, so we output json again

plumpy:
cc @karlkfi @sikevux

Revert the change in #241. This isn't a clean revert because #243 restructured things so much, but this puts things back the way they were.

@karlkfi noticed in #243 that if your jsonpath expression returns a complex (non-string) type, instead of JSON, it outputs a bunch of golang-formatted strings, like `map[EXECUTE:[admin-group] READ:[admin-group user-group] WRITE:[admin-group]]`.

### DIFF
--- a/cmd/output/output.go
+++ b/cmd/output/output.go
@@ -70,13 +70,20 @@ func MarshalToJsonPathWrapper(expression string) OutputFormater {
 			return nil, fmt.Errorf("Failed to parse jsonpath expression: %v", err)
 		}
 
-		buffer := new(bytes.Buffer)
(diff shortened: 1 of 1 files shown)


karlkfi:
I would love to see some tests around this to verify that the output remains consistent or only gets changes intentionally. But I guess that's what the revert is trying to accomplish. So I'd rather see the revert first. Maybe tests can be a follow up PR? Otherwise we can never really change it safely.

plumpy:
Yeah, I 100% agree with that, but I guess I just wanted to put things back how they were and then move forward from there...

PR #247, opened 2020-05-04
chore(gateapi): Update gate api and add gateapi-update.sh

karlkfi:
I was going to add support for returning the execution ID from spin pipeline execute, but it turns out that feature hasn't made it into the swagger gateapi here yet. So I tried to update the gateapi, which turned out to be a nightmare without java installed locally and a full gate build environment. So I wrote a `Dockerfile.swagger` and an `gateapi-update.sh` to automate it.

What I still can't figure out is how the Gate API is versioned or how we decide which version of the Gate API to put into spin. The last commit in the gateapi (other than mine running goimports) was to update it to 1.18.0, but I don't even see a v1.18.0 tag in gate...

The latest "version-" tag is `version-1.15.2`.
The latest "v" tag is `v6.42.0`.
The latest "1." tag is `1.221`.
The latest "release-" branch is `release-1.20.x`.

How is this supposed to work?

Once I know that I can update the script to use the right one and regenerate new api.


karlkfi:
FYI, in order to get tests to pass, I had to re-order the expected test output from `spin account get` again. It has to be the same order as the fields in swagger.yaml, even tho they seem randomly ordered at generation time. Seems like a swagger bug.

danielpeach:
re: your versioning question - I think (though not 100% sure) that ideally `spin@HEAD` on `master` is kept up to date with `gate@HEAD`. This repo has release branches that are cut with the rest of the Spinnaker microservices, so `spin` on `release-1.20.x` should work with `gate` on `release-1.20.x`.

re: this PR, I like the direction, though it would be really great if we had a GitHub action that 

* polled Gate repo
* generated swagger API on code changes
* ran the `swaggerapi/swagger-codegen-cli` image and generated the go code
* committed those changes if possible

kevinawoo:
hey @karlkfi, i updated the spincli to use spinnaker release 1.21.0
with https://github.com/spinnaker/spin/pull/254

Can you close this PR if it addresses your issue?

kevinawoo:
goign to close this PR for now. Please open a new issue or another PR if you find something. Cheers!

kevinawoo:
I took a second look and fixed it up using docker like you have and updated it with the latest Gate `master` branch.

As for versioning, `spin` master should follow `Gate` master.

It's only when the release branch is cut and a release is made, do the versions matter.

LMK if https://github.com/spinnaker/spin/commit/2b364bb5214e521a74faed8723341a2164a090d0 doesn't address your concerts, we can open another PR or an issue.

PR #250, opened 2020-06-19
chore(gateapi): update gate api to 1.20

perek:
Upgrade gate api to pull in new methods, and this fix PR: https://github.com/spinnaker/gate/pull/1258

This change does not affect the methods used in the spin pipeline command, build still passes.


perek:
@ethanfrogers

PR #253, opened 2020-06-30
test(account/get): use jsondiff to compare responses

kevinawoo:
gateapi 1.21.x autogenerated code from swagger will return json payload with the keys sorted differently. This is to address that issue

### DIFF
--- a/cmd/account/get_test.go
+++ b/cmd/account/get_test.go
@@ -17,6 +17,7 @@ package account
 import (
 	"bytes"
 	"fmt"
+	"github.com/nsf/jsondiff"
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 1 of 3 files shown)


karlkfi:
This changes the test to ignore ordering. Was that intentional?
It seems like the order of the API responses comes from the order in the Schema, but the schema struct in spin is generated (i think) by iterating over a Go map, which causes randomized ordering if the map is over a certain size.
Wouldnt it be better to strictly sort the json responses for consistency, rather than ignoring the order?

kevinawoo:
From https://www.json.org/json-en.html
> In JSON, they take on these forms:
>
> An object is an unordered set of name/value pairs. An object begins with {left brace and ends with }right brace. Each name is followed by :colon and the name/value pairs are separated by ,comma.

Since it's an unordered list, I don't think there's any value in sorting it. Personally, I get a bit weirded out when I see something sort `id` or some other key I expect up top, to be put in the middle somewhere.

However there is some value of having it ordered so you can `diff` using some plaintext diffing tool.

@karlkfi If you want to make a issue or an PR to sort all the output, 👍. We can discuss further on that issue/PR. Don't forget to sort all the nested keys also!

kevinawoo:
@mergifyio backport release-1.21.x

PR #255, opened 2020-07-07
fix(pipelineSave) Fixed issue with pipeline save (1.21.x)

gsapkal:
With latest spinnaker version 1.21.x The ApplicationControllerApi.GetPipelineConfigUsingGET returns 404 for non-existing pipeline. 
Updated pipeline save to check for status http.StatusOK and foundPipeline both.


gsapkal:
Will create new PR with correct format.

PR #263, opened 2020-07-19
fix(gatewayClient): retrieve gate endpoint from struct method

hugocortes:
issue: https://github.com/spinnaker/spinnaker/issues/5913

I was unable to authenticate due this line: https://github.com/spinnaker/spin/blob/master/cmd/gateclient/client.go#L134 which was due to `gateEndpoint` being an empty string value as I provide the value through `~/.spin/config` and not as cli arg.

This fix simply calls struct method `GateEndpoint` which returns arg value or the config value.

### DIFF
--- a/cmd/gateclient/client.go
+++ b/cmd/gateclient/client.go
@@ -130,7 +130,7 @@ func NewGateClient(ui output.Ui, gateEndpoint, defaultHeaders, configLocation st
 
 	gateClient.httpClient = httpClient
 
-	updatedConfig, err := authenticateOAuth2(ui.Output, httpClient, gateEndpoint, gateClient.Config.Auth)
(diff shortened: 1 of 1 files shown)


Chinikins:
I tested this and it worked for me. I also noticed that the updating config checks other auth mechanisms that aren't in the config originally which ends up writing the ldap null to your config. I would be curious if you could try out this commit: https://github.com/spinnaker-hackathon/spin/commit/6a766f47f6c3f87ff7aee96f225eb1f3226e8b0f as I think that might be a cleaner way to do the config update.

kevinawoo:
https://github.com/spinnaker/spin/pull/271 should fix this, going to close it. Feel free to open a new PR if it doesn't work for you!

PR #264, opened 2020-07-19
return execution id on pipeline execute command

suhrud-kumar:
We are building a wrapper around spin cli, so if spin cli could execution id then we could use it to poll spinnaker to get the status of the pipeline. Looks like the gate api already returns the pipeline execution id but we are not returning it to user. I made a couple of changes for this to work.

before code changes
```
spin pipeline execute -a testapp -n samplemicroservice
Pipeline execution started
```

after
```
./spin pipeline execute -a testapp -n samplemicroservice
{
 "ref": "/pipelines/01EDK74BBVZ97QPWDXBYWARVH4"
}
```

I understand the code in gateapi is auto-generated using swagger, so probably something needs to be updated in gate service as well (probably here - https://github.com/spinnaker/gate/blob/master/gate-web/src/main/groovy/com/netflix/spinnaker/gate/controllers/PipelineController.groovy#L198). I'm not well versed with java so I'm not sure what needs to be changed there. I'm updating my code here, so that we know what needs to be changed.

This also fixes - https://github.com/spinnaker/spinnaker/issues/5657

cc: @kevinawoo

### DIFF
--- a/cmd/pipeline/execute.go
+++ b/cmd/pipeline/execute.go
@@ -87,7 +87,7 @@ func executePipeline(cmd *cobra.Command, options *executeOptions) error {
 		}
 	}
 
-	resp, err := options.GateClient.PipelineControllerApi.InvokePipelineConfigUsingPOST1(options.GateClient.Context,
+	successPayload, resp, err := options.GateClient.PipelineControllerApi.InvokePipelineConfigUsingPOST1(options.GateClient.Context,
 		options.application,
 		options.name,
 		map[string]interface{}{"trigger": trigger})
@@ -100,7 +100,7 @@ func executePipeline(cmd *cobra.Command, options *executeOptions) error {
 		return fmt.Errorf("Encountered an error executing pipeline, status code: %d\n", resp.StatusCode)
 	}
 
-	options.Ui.Success("Pipeline execution started")
+	options.Ui.JsonOutput(successPayload)
 
 	return nil
 }
--- a/gateapi/pipeline_controller_api.go
+++ b/gateapi/pipeline_controller_api.go
@@ -578,12 +578,13 @@ func (a *PipelineControllerApiService) GetPipelineUsingGET(ctx context.Context, 
  @param optional (nil or map[string]interface{}) with one or more of:
      @param "trigger" (interface{}) trigger
  @return */
-func (a *PipelineControllerApiService) InvokePipelineConfigUsingPOST1(ctx context.Context, application string, pipelineNameOrId string, localVarOptionals map[string]interface{}) ( *http.Response, error) {
+func (a *PipelineControllerApiService) InvokePipelineConfigUsingPOST1(ctx context.Context, application string, pipelineNameOrId string, localVarOptionals map[string]interface{}) (interface{}, *http.Response, error) {
 	var (
 		localVarHttpMethod = strings.ToUpper("Post")
-		localVarPostBody interface{}
-		localVarFileName string
-		localVarFileBytes []byte
+		localVarPostBody   interface{}
+		localVarFileName   string
+		localVarFileBytes  []byte
+		successPayload     interface{}
 	)
 
 	// create path and map variables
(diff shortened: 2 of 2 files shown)


kevinawoo:
hey @suhrud-kumar you're right that we need to have the fix in gate. 

I'm hesitant to merge this because it's going to get dropped in the next release because we just replace the autogenerated code blindly.

kevinawoo (changes requested):
hey @suhrud-kumar you're right that we need to have the fix in gate.

I'm hesitant to merge this because it's going to get dropped in the next release because we just replace the autogenerated code blindly.

suhrud-kumar:
> hey @suhrud-kumar you're right that we need to have the fix in gate.
> 
> I'm hesitant to merge this because it's going to get dropped in the next release because we just replace the autogenerated code blindly.

yes, I agree. I just updated code here so that it could serve as pointer for what needs to be fixed in gate. Do you want me to open an issue with gate for this ?

kevinawoo:
Yeah, you can open a issue in [spinnaker/spinnaker](https://github.com/spinnaker/spinnaker/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc) to update it and someone else might take it.

Chinikins:
I don't think this is fixed in the gate yet but I think my PR could help here too: https://github.com/spinnaker/spin/pull/311 we often need to get the execution IDs for pipelines for various reasons too.

boonware:
Any update on this? Looks like Spin CLI still does not support this. It's a crucial feature.

Chinikins:
could we update this PR with latest master branch? :pray: Just being able to see the response returned I think would be helpful and enables users to parse it as needed.

girish-netskope:
Any updates on this? Its a cruical thing to have.

cmd/pipeline/execute.go
```diff
@@ -87,7 +87,7 @@ func executePipeline(cmd *cobra.Command, options *executeOptions) error {
 		}
 	}
 
-	resp, err := options.GateClient.PipelineControllerApi.InvokePipelineConfigUsingPOST1(options.GateClient.Context,
+	successPayload, resp, err := options.GateClient.PipelineControllerApi.InvokePipelineConfigUsingPOST1(options.GateClient.Context,
```
karlskewes, line 90: Instead of doing the JSON decoding in the generated file, why not attempt a JSON decode of the response body `resp` here?

gateapi/pipeline_controller_api.go
```diff
@@ -621,20 +621,24 @@ func (a *PipelineControllerApiService) InvokePipelineConfigUsingPOST1(ctx contex
...
 	if localVarHttpResponse.StatusCode >= 300 {
 		bodyBytes, _ := ioutil.ReadAll(localVarHttpResponse.Body)
-		return localVarHttpResponse, reportError("Status: %v, Body: %s", localVarHttpResponse.Status, bodyBytes)
+		return successPayload, localVarHttpResponse, reportError("Status: %v, Body: %s", localVarHttpResponse.Status, bodyBytes)
 	}
 
-	return localVarHttpResponse, err
+	if err = json.NewDecoder(localVarHttpResponse.Body).Decode(&successPayload); err != nil {
+		return successPayload, localVarHttpResponse, err
+	}
```
karlskewes, line 639: I think this can be moved to `execute.go` per my comment above.
Then do a pass to ensure errors are checked and `resp.Body` is closed, etc.

karlskewes (changes requested):
This PR looks pretty close on initial inspection. 
Was there a reason to do the JSON decoding in the swagger generated code instead of in `execute.go` ?

cmd/pipeline/execute.go
karlskewes: It looks like it might be better to use the newer Gate API `options.GateClient.PipelineControllerApi.InvokePipelineConfigViaEchoUsingPOST` instead.

This returns an `interface{}` which we can marshal into a struct.
The struct will hold a list of executed pipelines because a single trigger (`spin`'s POST) can invoke multiple pipelines.

See Gate groovy returned object which would land in the Go empty interface.
```
    return [
      eventId: eventId,
      ref    : String.format("/pipelines/%s", executionId)
    ]
```
Per: https://github.com/spinnaker/gate/blob/6e751aa7d7c9a4f116b85b6f4b73d5203d902ce7/gate-web/src/main/groovy/com/netflix/spinnaker/gate/services/PipelineService.groovy#L94-L116

shufanhao:
shall we merge this PR ?

cmd/pipeline/execute.go
shufanhao: what's the different of `InvokePipelineConfigViaEchoUsingPOST` and `InvokePipelineConfigUsingPOST1` ?

PR #266, opened 2020-07-20
feat(client): add error output to auth

KeisukeYamashita:
## What

* Attached `err` for UI output auth error

## Why

It is hard to know why it failed.


ezimanyi:
@KeisukeYamashita : Looks like there's now a conflict, but this should be good to merge once that is addressed!

KeisukeYamashita:
@ezimanyi Thank you for your review. I've solved the conflicts. Please take a look when you have time 🙇

PR #268, opened 2020-07-20
fix(gateclient): broken gate endpoint

KeisukeYamashita:
## What

Fix broken Gate endpoint configuration API.

## Why

The current version only reads Gate endpoint from the CLI argument and doesn't refer to the spin config`s `.gate.endpoint`.


kevinawoo:
https://github.com/spinnaker/spin/pull/271 should fix this, going to close it. Feel free to open a new PR if it doesn't work for you!

PR #269, opened 2020-07-20
fix(config): broken spinconfig

KeisukeYamashita:
## What

The spin config is broken and it has no backward compatibility.  Also ran `go mod tidy`.

In current latest master version, the spin requires spin config like below:

```yaml
Auth:
    Basic: null
    Enabled: true
    GoogleServiceAccount: null
    Iap: null
    IgnoreCertErrors: false
    Ldap: null
  OAuth2:
      AuthUrl: https://xxx.com
      CachedToken:
       access_token: MASKED
        expiry: "0001-01-01T00:00:00Z"
        token_type: bearer
      ClientId: MASKED
      ClientSecret: MASKED
      Scopes:
      - openid email
      TokenUrl: https://yyy.com
    X509: null
 Gate:
   Endpoint:                                         
```

## Why

This is because spin used to use `sigs.k8s.io`'s YAML util packages.
In this package, for example, the marshal is done by first marshaling it as JSON and then marshal again to YAML.

https://github.com/kubernetes-sigs/yaml/blob/9fc95527decd95bb9d28cc2eab08179b2d0f6971/yaml.go#L15-L28

Therefore, the YAML annotations will be omitted. That's why the current spin requires and updates the spin config exactly the same as Go struct.

I've changed to use the Go native YAML package because we don't need to marshal/unmarshal to JSON

### DIFF
--- a/cmd/gateclient/client.go
+++ b/cmd/gateclient/client.go
@@ -46,7 +46,7 @@ import (
 	"golang.org/x/crypto/ssh/terminal"
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
-	"sigs.k8s.io/yaml"
+	"gopkg.in/yaml.v2"
 )
 
 const (
@@ -221,7 +221,7 @@ func userConfig(gateClient *GatewayClient, configLocation string) error {
 
 	yamlFile, err := ioutil.ReadFile(gateClient.configLocation)
 	if yamlFile != nil {
-		err = yaml.UnmarshalStrict([]byte(os.ExpandEnv(string(yamlFile))), &gateClient.Config)
+		err = yaml.Unmarshal([]byte(os.ExpandEnv(string(yamlFile))), &gateClient.Config)
 		if err != nil {
 			gateClient.ui.Error(fmt.Sprintf("Could not deserialize config file with contents: %s, failing.", yamlFile))
 			return err
@@ -229,6 +229,8 @@ func userConfig(gateClient *GatewayClient, configLocation string) error {
 	} else {
 		gateClient.Config = config.Config{}
 	}
+
+	fmt.Println(gateClient.Config)
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


kevinawoo:
Hey @KeisukeYamashita going to close this in favor of #262.

I think it makes more sense to be more strict rather than not as strict.

Here's the comments from that PR that's in [master](https://github.com/spinnaker/spin/blob/master/cmd/gateclient/client.go#L229-L233) now:
> // Please note that https://github.com/spinnaker/spin/pull/243 introduced better coding standards and
> // as a result, your auth config needs to match the yaml tags through all the config structs
> // e.g. the yaml tags for oauth in the config are set in the local oauth package here
> // but unmarshal to an upstream oauth package, so the cached token needs to match
> // https://godoc.org/golang.org/x/oauth2#Token

Closing this, but feel free to open up a new issue or comment on https://github.com/spinnaker/spinnaker/issues/5913 if this doesn't actually fix it the way you'd like it to.

KeisukeYamashita:
@kevinawoo  So does the spin works with the spinconfig in the example? As this PR says, the Marshal and Unmarshal isn't working...

kevinawoo:
Can you try fixing the indents in your config to see if that fixes it?

Some indents widths I see are 1, 2 and 4 spaces wide.

KeisukeYamashita:
@kevinawoo With the latest master, `spin` will overwrite a `spin` config like below. Could you reopen this PR please?

```yaml
Auth:
  Basic: null
  Enabled: true
  GoogleServiceAccount: null
  Iap: null
  IgnoreCertErrors: false
  Ldap: null
  OAuth2:
    AuthUrl: https://github.com/login/oauth/authorize
    CachedToken:
      access_token: xxxxxxxxxxxxxxxx
      expiry: "0001-01-01T00:00:00Z"
      token_type: bearer
    ClientId: xxxxxxxxxxxxxxxx
    ClientSecret: xxxxxxxxxxxxxxx
    Scopes:
    - openid email
    TokenUrl: https://github.com/login/oauth/access_token
  X509: null
Gate:
  Endpoint: https://xxx.com
```

Chinikins:
Hey @KeisukeYamashita could you elaborate on the error? Was the scopes part originally written as:

 Scopes:
    - openid
    - email

and it gets rewritten to one dash? That's just me guessing what the issue is, the actual error you're getting would be helpful, thanks!

PR #274, opened 2020-07-21
Fix(cli): Lint issues for project command

jan25:
Changes based on:
* use `const` for constants
* `go fmt`


cmd/project/project.go
```diff
@@ -35,16 +35,16 @@ func NewProjectCmd(rootOptions *cmd.RootOptions) *cobra.Command {
 	}
 	cmd := &cobra.Command{
 		Use:     "project",
-		Aliases: []string{"project", "prj"},
+		Aliases: []string{"prj"},
```
Chinikins, line 38: I think this might break being able to call `spin project <sub-commands>` and makes it just `spin prj` but not sure

jan25: The `Use:` bit handles full command name. I removed it because it was duplicated in help text.

ezimanyi:
I'll fix the minor conflict with #270 mostly so this doesn't stay outstanding for too long, as it touches a lot of lines and might end up with more conflicts.

PR #275, opened 2020-07-22
Fix(cli): Use yaml as default output format

jan25:
Because `json` can be too verbose for cli output. If needed it is still available with `-o json` anyway.


jan25:
Please let me know if this makes sense. I'll update the failing tests after your review :)

jan25:
Closed because this change may break other tools using default json output.

PR #276, opened 2020-07-22
Fix(CI): setup linters

jan25:
Based on https://github.com/spinnaker/spin/pull/274#pullrequestreview-452901257.

Added new CI job for linting because it can't be run in same job as build(`go get` brings in dependencies for linting. See https://github.com/golangci/golangci-lint-action/issues/23#issuecomment-647177770)

This change will make sure code is properly formatted in future commits.

Fixed violations from `gofmt`, `goimports` and `govet`. Left a `TODO` to fix violations from `golint` which can be fixed separately, as automatic fix is not possible and will have to be done manual.

All non-config changes in this PR are done with `golangci-lint run --fix`


ajordens:
LGTM

Don't know the code-base at all but changes look to be aesthetic-only.

PR #278, opened 2020-07-23
feat(execute): add individual parameter option

ethanfrogers:
adds an option to specify multiple key=value parameters instead of using
a parameter file. the default behavior is kept the same but if the user
passes `-p` instead, it will parse the arguments as key/value pairs.

### DIFF
--- a/cmd/pipeline/execute.go
+++ b/cmd/pipeline/execute.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/spf13/cobra"
 
@@ -31,6 +32,7 @@ type executeOptions struct {
 	name          string
 	parameterFile string
 	artifactsFile string
+	parameters    []string
 }
 
 var (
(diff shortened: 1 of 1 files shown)


ethanfrogers:
Example usage:
```
$ spin pipeline execute --application app --name pipeline -p PARAM1=foo --parameter PARAM2=bar
```

jan25:
My suggestion here - instead of having `-p` for each pair could we support all pairs in same argument?
E.g.` -p KEY1=VAL1;KEY2=VAL2`

Ideas?

ethanfrogers:
@jan25 - thanks for the suggestion! What do you think would make that easier than repeated `-p` arguments? My reasoning for this approach was that it seems like a pretty common pattern among other CLI tools like `kubectl`.

jan25:
Just something that seems reasonable while typing in the arguments for e.g. all pairs will be in one place as a single string. If its very common to type/allow repeated flags this PR looks fine :)

ezimanyi:
I could be wrong `pflag` library may actually allow either `-p a=b,c=d` or `-p a=b -p c=d` (or some combination)...worth checking this!

ethanfrogers:
Confirmed! `StringSliceVarP` accepts arguments in both forms - `-p a=b,b=c` or  `-p a=b -p b=c` [Docs for the curious](https://pkg.go.dev/github.com/spf13/pflag?tab=doc#StringSliceVar).

Chinikins:
I opened a PR to document this change as well as some of the other changes added to the Spin CLI but I'm unsure how the documentation coordinates with Spin CLI release cycles.

ezimanyi:
Right now we don't have versioned documentation, but that is on the [roadmap](https://github.com/spinnaker/governance/blob/master/rfc/website-revamp.md#goals) for the documentation SIG.  At this point we either note in the docs PR what version is needed when describing the feature, or wait until the feature is released to merge the docs change.

PR #279, opened 2020-07-23
chore(gateclient): use os.UserHomeDir instead of mitchellh/go-homedir

micnncim:
Now that Go 1.12 provides `os.UserHomeDir`, this PR uses it instead of [mitchellh/go-homedir](https://github.com/mitchellh/go-homedir). Thanks to this, we can reduce third-party dependency.

## Refenrences

- https://golang.org/doc/go1.12#os

### DIFF
--- a/cmd/gateclient/client.go
+++ b/cmd/gateclient/client.go
@@ -30,12 +30,10 @@ import (
 	_ "net/http/pprof"
 	"net/url"
 	"os"
-	"os/user"
 	"path/filepath"
 	"strings"
 	"syscall"
 
-	"github.com/mitchellh/go-homedir"
 	"github.com/pkg/errors"
 	"golang.org/x/crypto/ssh/terminal"
 	"golang.org/x/oauth2"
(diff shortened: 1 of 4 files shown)


cmd/gateclient/client.go
```diff
@@ -606,3 +595,15 @@ func securePrompt(output func(string), inputMsg string) string {
 	secret := string(byteSecret)
 	return strings.TrimSpace(secret)
 }
+
+func expandFilepath(path string) (string, error) {
```
Chinikins, line 599: this might make more sense under the util package than under the gate client package?

micnncim: I agree with you. I’ll update it!

micnncim:
@ezimanyi @Chinikins I've updated the function. Could you confirm this?

ezimanyi:
@micnncim : Looks good!  Looks like you will need to merge in `master` but it should auto-merge as soon as this is done.

micnncim:
@ezimanyi I'm not sure why mergify doesn't work... Do you know something?

Chinikins:
Rebase it with latest master and it should merge!

micnncim:
Oh, I misunderstood something :bow: Thank you, I've rebased it.

PR #283, opened 2020-07-28
feat(golang-cli): ignore swagger generated code in gateapi/

kevinawoo:
the code in `/gateapi` is autogenerated by swagger. There's not a really good reason to enforce any kind of linting in there.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,5 +1,7 @@
 run:
   modules-download-mode: readonly
+  skip-dirs:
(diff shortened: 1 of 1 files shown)


kevinawoo:
@jan25 can you take a look? 🙏

jan25:
@kevinawoo looks good! I would expect the generated code to follow the conventions anyway :) But makes sense to ignore it since we barely read it.

PR #286, opened 2020-07-29
test(jsonpath): add tests for non-primitives

sikevux:
This change adds tests to prevent unintentional breakage related to
output of non-primitive jsonpath outputs.

The changes made in https://github.com/spinnaker/spin/pull/241 do not
pass these tests, as they return Go-syntax.

Updating client-go to v0.19.0-rc.2 that contains the fix for printing
JSON still does not pass the tests as the output would be encapsulated
in `[]`

### DIFF
--- a/cmd/application/get_test.go
+++ b/cmd/application/get_test.go
@@ -55,6 +55,28 @@ func TestApplicationGet_json(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


sikevux:
Rebased my changes on current master for easier merge 😄

PR #291, opened 2020-08-06
fix(projectSave) if pipeline id is sent in json, set project id to that

Chinikins:
Addressing https://github.com/spinnaker/spinnaker/issues/3651#issuecomment-665491856 -> if project id is sent, the project will be saved with that ID sent to it.


Chinikins:
Upon evaluating this some more, all this was "fixing" was a bad user flow of if spinnaker had set a project id and the user then wanted to set it to their own later. Without it, the id is either continually set by the user from first call or either continually set by spinnaker from first call which makes more sense to me.

PR #294, opened 2020-08-18
fix(pipelineTemplates): catch errors from save/create task

Chinikins:
This updates the gate to master to address https://github.com/spinnaker/spinnaker/issues/5967 and checks the task response returned to see if pipeline templates are actually saved correctly or not. Gate API was updated to b3ede47f to fix this bug as well.

### DIFF
--- a/cmd/pipeline-template/save.go
+++ b/cmd/pipeline-template/save.go
@@ -123,11 +123,13 @@ func savePipelineTemplate(cmd *cobra.Command, options *saveOptions) error {
 			saveResp.StatusCode)
 	}
 
-	taskSucceeded := orca_tasks.TaskSucceeded(saveRet)
-	if taskSucceeded {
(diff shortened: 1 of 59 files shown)


cmd/pipeline-template/save.go
```diff
@@ -121,6 +123,13 @@ func savePipelineTemplate(cmd *cobra.Command, options *saveOptions) error {
 			saveResp.StatusCode)
 	}
 
+	if len(saveRet) > 0 {
+		taskSucceeded := orca_tasks.TaskSucceeded(saveRet)
```
kevinawoo, line 127: Thoughts on splitting up the PR because this PR is doing 2 things at once? 

`pipelineTemplate status and gate task tracking` feature along with `updating gateapi`

kevinawoo: talked to @Chinikins, it's fine to combine time together

PR #298, opened 2020-08-28
fix(pipelineTemplateSave): fix returning error on existing pipeline template saving

Chinikins:
See issue: https://github.com/spinnaker/spinnaker/issues/6008

Recent changes to pipeline template saving did not take into account saving existing pipeline templates and will return an error without the status check fixed.

### DIFF
--- a/cmd/pipeline-template/save.go
+++ b/cmd/pipeline-template/save.go
@@ -117,7 +117,7 @@ func savePipelineTemplate(cmd *cobra.Command, options *saveOptions) error {
 		return saveErr
 	}
 
-	if saveResp.StatusCode != http.StatusAccepted {
+	if saveResp.StatusCode != http.StatusAccepted && saveResp.StatusCode != http.StatusOK {
 		return fmt.Errorf("Encountered an error saving pipeline template %v, status code: %d\n",
 			templateJson,
 			saveResp.StatusCode)


Chinikins:
@spinnakerbot add-label backport-candidate
Adding the label because this is a regression that got introduced in 1.22. It doesn't have test, but I feel like there's not a whole lot of value to add a test for this condition. I'll leave it up to the release-manager to decide.

german-muzquiz:
If possible I'd prefer to have test coverage for the regression. When there are no tests we may introduce another regression in the future if the response code ever changes again.

Here for example, what happens if gate changes its response to 204 (no content) or 201 (created)? Both seem plausible response codes after saving a pipeline. So I think the code at least needs to check for 2xx response codes in general to verify if a request was successful.

Chinikins:
hi @german-muzquiz  - valid request and there currently is a test, but I believe the test might be broken here: https://github.com/spinnaker/spin/blob/master/cmd/pipeline-template/save_test.go#L312 where the mux needs to change to say `http.StatusAccepted` to `http.StatusOk`. I'm not 100% sure though which is why I didn't change the existing test.

Overall I think the testing framework needs to change to be able to take the latest from gate and have started some work on it but it isn't ready for PR as it's a large change.

I could make another PR to update the test, would just appreciate some review to make sure it's the right change to the test. :)

Chinikins:
I've created a PR for to fix the test, I feel more certain of it now after doing some testing, please review: https://github.com/spinnaker/spin/pull/301

german-muzquiz:
@Mergifyio backport release-1.22.x

PR #300, opened 2020-09-01
chrore(gateapi): go format client

KeisukeYamashita:
## What

As title

## Why 

Style is broken.


Chinikins:
These files get autogenerated via swagger. I don't know if there's a way to run the swagger cli command that would also do a go fmt on the files, but the change should probably be done with that flow instead of changing these files directly.

KeisukeYamashita:
Could you please tell me what commands you executed? Perhaps we can modify it so that it can be formatted.

PR #301, opened 2020-09-02
fix(pipelineTemplateSave): fix tests for saving pipeline template

Chinikins:
This is to fix the test for the issues mentioned in: https://github.com/spinnaker/spin/pull/298

This addresses github issues https://github.com/spinnaker/spinnaker/issues/6008 and part of https://github.com/spinnaker/spinnaker/issues/6003#issuecomment-685791703

I've confirmed that the test fails properly without the fix in PR 298 and passes correctly with the fix in.


kevinawoo (approved):
lgtm

german-muzquiz:
@Mergifyio backport release-1.22.x

PR #304, opened 2020-09-10
feat(azureOauth2) Updated oauth2 auth for azure support

gsapkal:
Updated the oauth2 to work with azure oauth2, added configuration support for Authcode options.
Fixed issue with api-gateway endpoint not read from config. 
Updated the auth workflow to open default browser and get act on callback to proceed. No need to copy past url and code from oauth2 redirect response.


gsapkal:
Messed with merge

PR #305, opened 2020-09-10
fix(azureOauth2): fixed oauth2 auth for azure custom config support

gsapkal:
Updated the oauth2 to work with azure oauth2 custom config, Added support for AuthCodeOptions options in config.
Updated the auth workflow to open default browser with auth url.  No need to copy past url and code from oauth2 redirect response.
If the default browser doesn't work this fallbacks to printing url on console with instruction.

### DIFF
--- a/cmd/gateclient/client.go
+++ b/cmd/gateclient/client.go
@@ -519,10 +519,15 @@ func login(httpClient *http.Client, endpoint string, accessToken string) error {
 		return err
 	}
 	loginReq.Header.Set("Authorization", fmt.Sprintf("Bearer %s", accessToken))
-	_, err = httpClient.Do(loginReq) // Login to establish session.
+	resp, err := httpClient.Do(loginReq) // Login to establish session.
 	if err != nil {
 		return errors.Errorf("login failed %s", err)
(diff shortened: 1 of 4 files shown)


ElenaForester:
Hi! When this PR will be merged? Looking forward for it!

Chinikins:
Hi, is there a linked bug that would describe the behavior you're fixing? Does your config file still look like for what we have for example oauth: https://github.com/spinnaker/spin/blob/master/config/example.yaml#L25? I don't have experience with spin CLI and authenticating to azure so a little more context could help the review process, thanks!

gsapkal:
> Hi, is there a linked bug that would describe the behavior you're fixing? Does your config file still look like for what we have for example oauth: https://github.com/spinnaker/spin/blob/master/config/example.yaml#L25? I don't have experience with spin CLI and authenticating to azure so a little more context could help the review process, thanks!

@Chinikins I have updated example config. This PR enables the advanced config for Oauth2. These are needed based on organizations oauth2 config.

Chinikins:
@gsapkal could you please resolve conflicts?

PR #308, opened 2020-11-11
chose(dep): Upgrade go version to 1.15

micnncim:
Upgraded the Go version to [`1.15`](https://golang.org/doc/go1.15) and fixed the use of different versions in CI.


micnncim:
@kevinawoo @Chinikins PTAL

kevinawoo (approved):

PR #309, opened 2020-11-11
refactor(cmd): Clean up value checks

micnncim:
## What

Removed unnecessary control flows and tidied up some blocks. The behavior wouldn't be changed.

## Why

To improve readability and maintainability and allow more users to easily contribute to this project.

PR #311, opened 2020-11-24
feat(applicationPipelines): ability to get execution IDs of pipelines under an application and cancel some or all with reasons

Chinikins:
This PR adds a pipeline subcommand to `spin application` with 3 subcommand options.

`spin application pipelines get -a <appName>` will list all executions of a pipeline under an application name.

You can narrow this down by passing the status of the pipeline you're looking for. i.e. `-s RUNNING` for example or `-s NOT_STARTED`

If you have the execution ID of a pipeline, you can do:

`spin application pipelines cancel -i <ID>`

If you want to cancel all the pipelines of a particular status under an application and pass a reason for it, you can run:

`spin application pipelines cancel-all -a <appName> -s RUNNING -r "killing zombie executions"`

If you don't pass the status, it will cancel all executions.

### DIFF
--- a/cmd/application/cancel_pipeline.go
+++ b/cmd/application/cancel_pipeline.go
@@ -1,0 +1,92 @@
+// Copyright (c) 2020, Anosua "Chini" Mukhopadhyay
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+
+package application
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+
+	"github.com/antihax/optional"
+	"github.com/spf13/cobra"
+
+	gate "github.com/spinnaker/spin/gateapi"
+)
+
+type cancelPipelineOptions struct {
+	*pipelineOptions
+	id     string
+	reason string
+}
+
+var (
+	cancelPipelinesShort   = "Cancel the pipeline for the specified pipeline execution id"
+	cancelPipelinesLong    = "Cancel the pipeline for the specified pipeline execution id"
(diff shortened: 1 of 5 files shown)


cmd/application/cancel_pipeline.go
```diff
@@ -0,0 +1,92 @@
...
+
+func cancelPipelineWithID(options *cancelPipelineOptions, id string) error {
+	if id == "" {
+		return errors.New("execution ID must be passed in")
+	}
+
+	pipeline, resp, err := options.GateClient.ApplicationControllerApi.CancelPipelineUsingPUT(options.GateClient.Context, id, &gate.ApplicationControllerApiCancelPipelineUsingPUTOpts{Reason: optional.NewString(options.reason)})
+
+	if resp != nil {
+		if resp.StatusCode == http.StatusNotFound {
```
kevinawoo, line 79: I just saw https://github.com/spinnaker/spin/pull/309, which cleans up some of the checking by using switch statements, it'd be nice if we're consistent 😍

cmd/application/get_pipelines.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func getPipelines(cmd *cobra.Command, options *getPipelinesOptions, args []string) ([]interface{}, error) {
+	if options.applicationName == "" {
+		return nil, errors.New("Application name must be passed in")
+	}
+
+	app, resp, err := options.GateClient.ApplicationControllerApi.GetPipelinesUsingGET(options.GateClient.Context, options.applicationName, &gate.ApplicationControllerApiGetPipelinesUsingGETOpts{Expand: optional.NewBool(options.expand), Statuses: optional.NewString(options.status)})
+	if resp != nil {
+		if resp.StatusCode == http.StatusNotFound {
```
kevinawoo, line 81: ditto https://github.com/spinnaker/spin/pull/311/files#r533745417

kevinawoo:
@Chinikins would you mind adding some tests? 🙏

cmd/application/cancel_pipeline.go
Chinikins: indeed, let me fix!

kevinawoo: @Chinikins was this was changed to use `switch case` statements?

Chinikins: @kevinawoo yes, pushed the change now

kevinawoo (approved):

PR #313, opened 2021-02-03
Change release tag follow semver

KeisukeYamashita:
## What

As described in [Tag Spin release versions with semver](https://github.com/spinnaker/spinnaker/issues/6189), in Go projects, it's better common to tag the release with semver.

This has many advantages, making it easy to import for users.


kevinawoo (approved):
lgtm

PR #320, opened 2021-05-26
chore(config): Ensure consistent marshalling/unmarshalling

vatine:
The sigs.k8s.io/yaml package round-trips YAML through JSON. This means that
without JSON tags on structs, they end up getting the default JSON naming
once marshalled. This change adds JSON tags to all structs, making sure that
the tags after writing the config file back is consistent with what the tags in the source code say.

This PR provides a fix for https://github.com/spinnaker/spinnaker/issues/6434

### DIFF
--- a/config/auth/authconfig.go
+++ b/config/auth/authconfig.go
@@ -25,13 +25,13 @@ import (
 
 // Config is the CLI's authentication configuration.
 type Config struct {
-	Enabled          bool           `yaml:"enabled"`
-	IgnoreCertErrors bool           `yaml:"ignoreCertErrors"`
-	X509             *x509.Config   `yaml:"x509,omitempty"`
-	OAuth2           *oauth2.Config `yaml:"oauth2,omitempty"`
-	Basic            *basic.Config  `yaml:"basic,omitempty"`
-	Iap              *config.Config `yaml:"iap,omitempty"`
-	Ldap             *ldap.Config   `yaml:"ldap,omitempty"`
+	Enabled          bool           `json:"enabled" yaml:"enabled"`
+	IgnoreCertErrors bool           `json:"ignoreCertErrors" yaml:"ignoreCertErrors"`
+	X509             *x509.Config   `json:"x509,omitempty" yaml:"x509,omitempty"`
+	OAuth2           *oauth2.Config `json:"oauth2,omitempty" yaml:"oauth2,omitempty"`
+	Basic            *basic.Config  `json:"basic,omitempty" yaml:"basic,omitempty"`
+	Iap              *config.Config `json:"iap,omitempty" yaml:"iap,omitempty"`
+	Ldap             *ldap.Config   `json:"ldap,omitempty" yaml:"ldap,omitempty"`
 
-	GoogleServiceAccount *gsa.Config `yaml:"google_service_account,omitempty"`
+	GoogleServiceAccount *gsa.Config `json:"google_service_account,omitempty" yaml:"google_service_account,omitempty"`
 }
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -1,0 +1,26 @@
+package config
+
+import (
+	"testing"
+
+	"sigs.k8s.io/yaml"
+)
+
+func TestBasicMarshalling(t *testing.T) {
+	var cfg Config
+
+	cfg.Gate.Endpoint = "test"
+
+	bs, err := yaml.Marshal(cfg)
+
+	if err != nil {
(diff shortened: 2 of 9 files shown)


dbyron-sf:
This looks great, thank you. What's involved in adding a test to demonstrate the need for this change?

vatine:
I guess the simplest would be to create a config struct, with minimal data in it, then marshal that to a []byte, then do a string comparison to an expected value.

vatine:
Minimal test case added that passes with the changes in the PR and fails without them.

config/config_test.go
```diff
@@ -0,0 +1,26 @@
...
+
+func TestBasicMarshalling(t *testing.T) {
+	var cfg Config
+
+	cfg.Gate.Endpoint = "test"
+
+	bs, err := yaml.Marshal(cfg)
+
+	if err != nil {
+		t.Errorf("Unexpected error unmarshalling YAML: %v", err)
```
dbyron-sf, line 17:
```suggestion
		t.Errorf("Unexpected error marshalling YAML: %v", err)
```

config/config_test.go
```diff
@@ -0,0 +1,26 @@
+package config
+
+import (
+	"testing"
+
+	"sigs.k8s.io/yaml"
+)
+
+func TestBasicMarshalling(t *testing.T) {
```
dbyron-sf, line 9: From https://github.com/spinnaker/spinnaker/issues/6434, I expected a test to start with yaml and get the same yaml back after a roundtrip.  I'd love someone who knows more go than me to chime in, and maybe this test is OK.  Maybe it's even better because it shows more specifically where the struggle is?

config/config_test.go
```diff
@@ -0,0 +1,26 @@
...
+	var cfg Config
+
+	cfg.Gate.Endpoint = "test"
+
+	bs, err := yaml.Marshal(cfg)
```
dbyron-sf, line 14: Maybe this is a go convention?  Perhaps a more descriptive variable name than `bs`?

config/config_test.go
vatine: Ooops, fixing.

config/config_test.go
vatine: In this specific case, it's short for "byte slice". Being a local variable and all, I don't mind what it's called, changing to `byteSlice`.

config/config_test.go
vatine: I am working on a second test that will try round-tripping.

config/config_test.go
dbyron-sf: sorry, my mind went straight to that other thing bs sometimes stands for :)

vatine:
@dbyron-sf Would it be possible to have another look at this?

dbyron-sf (approved):
LGTM.  Sorry for the delay.  I think we'll need some extra help getting this merged though, as I don't have privileges for that.  Maybe @Chinikins can help?

vatine:
@dbyron-sf No worries, I just realised it had hit 2+ weeks since the last inetraction and thought "maybe a gentle prod is appropriate".

vatine:
This PR has been approved for a while. Would it be possible for someone with merge privileges to chime in if there's something else needed?

Chinikins (approved):

config/auth/authconfig.go
```diff
@@ -25,13 +25,13 @@ import (
...
+	Enabled          bool           `json:"enabled" yaml:"enabled"`
+	IgnoreCertErrors bool           `json:"ignoreCertErrors" yaml:"ignoreCertErrors"`
+	X509             *x509.Config   `json:"x509,omitempty" yaml:"x509,omitempty"`
+	OAuth2           *oauth2.Config `json:"oauth2,omitempty" yaml:"oauth2,omitempty"`
+	Basic            *basic.Config  `json:"basic,omitempty" yaml:"basic,omitempty"`
+	Iap              *config.Config `json:"iap,omitempty" yaml:"iap,omitempty"`
+	Ldap             *ldap.Config   `json:"ldap,omitempty" yaml:"ldap,omitempty"`
 
-	GoogleServiceAccount *gsa.Config `yaml:"google_service_account,omitempty"`
+	GoogleServiceAccount *gsa.Config `json:"google_service_account,omitempty" yaml:"google_service_account,omitempty"`
```
dmivankov, line 36: note: this change requires manual patching of pre-existing ~/.spin/config file as now `GoogleServiceAccount: null` doesn't parse

PR #325, opened 2022-02-17
refactor(auth): improve authentication flow for oauth 2.0 with local webserver

GennadySpb:
Hello there! 

PR is designed to minimize the use of interactive mode for OAuth 2.0 authentication.

The base for the extension is inspired by https://github.com/youtube/api-samples/blob/master/go/oauth2.go


GennadySpb:
Hey @dbyron-sf ! Could your check out PR?

dbyron-sf:
> Hey @dbyron-sf ! Could your check out PR?

I suspect @kskewes-sf is in a better position to review this.

kskewes-sf:
Sure, I remember the interactive cli driven flow and this looks interesting.
I need to refresh myself on spin's oauth flow and get a local setup going which I want to do after we get release 1.27.0 out. Sound okay @GennadySpb ?
There are several reports of `spin` CLI being broken since 1.14 so it would be good to validate before introducing new changes.
That said will do a quick pass through PR.

cmd/gateclient/oauth.go
```diff
@@ -0,0 +1,85 @@
...
+
+func startWebServer() (codeCh chan string, err error) {
+	listener, err := net.Listen("tcp", localWebServer)
+	if err != nil {
+		return nil, err
+	}
+	codeCh = make(chan string)
+
+	go http.Serve(listener, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// TODO: add handle of `error`
```
kskewes-sf, line 25: This is commented but new in this PR. Can you give an example of what `error` is here?
Could be something for a follow up PR.

cmd/gateclient/oauth.go
```diff
@@ -0,0 +1,85 @@
...
+	"net/http"
+	"os/exec"
+	"runtime"
+
+	"golang.org/x/oauth2"
+)
+
+const localWebServer = "localhost:8085"
+
+func startWebServer() (codeCh chan string, err error) {
```
kskewes-sf, line 17: Can you add a test for this please? - see: https://pkg.go.dev/net/http/httptest
We can validate:
1. `code` FormValue is supplied in test and plumbed through to codeCh
2. http response recorder has correct output
3. listener closed
4. `error` FormValue (per next comment - if part of this PR).

kskewes-sf (changes requested):

cmd/gateclient/oauth.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Unable to retrieve token %v", err)
+	}
+	return tok, nil
+}
+
+// openURL opens a browser window to the specified location.
+// This code originally appeared at:
+//   http://stackoverflow.com/questions/10377243/how-can-i-launch-a-process-that-is-not-a-file-in-go
+func openURL(url string) error {
```
kskewes-sf, line 72: Using a local web server or prompt is optional in the linked code and this seems quite robust. If local web server doesn't work then use the prompt. https://github.com/youtube/api-samples/blob/master/go/oauth2.go

I'm not sure what the failure modes are for the various `openURL()` commands, like `exec` hanging with no feedback. I've managed to misconfigure Linux for `xdg-open` more than once :sweat_smile: 

If an error is returned we bail out at L48 with `log.Fatalf` and there is no way for a user to progress unless they fix the exec issue.

I wonder if we might want to do something like:
a) fall back to prompt for code?
b) make `localWebServer` selectable via spin config (default enabled?) similer to https://github.com/youtube/api-samples/blob/07263305b59a7c3275bc7e925f9ce6cabf774022/go/oauth2.go#L44 

WDYT?

PR #326, opened 2022-03-04
Support ignoring redirects during oauth2 flow

robert-blackman:
Hello, we came across an issue with the OAuth2 login flow. 

During a typical login, gate will respond with a 302 redirecting to deck. In the case that the predefined redirect URL is undefined (in our case it _is_ undefined) `localhost:8085` is set as the location. Due to the default behaviour of the [go http client](https://go.dev/src/net/http/client.go), this redirection is followed causing an error when the connection is inevitably refused.

This PR adds support for an `ignoreRedirects` configuration and/or `--ignore-redirects` command line arg that instructs the client not to follow redirects, the default behaviour should remain the same however. Additionally, the error message has been improved to help diagnose further failures.

Thanks!


dbyron-sf:
@mergifyio update

dbyron-sf:
@robert-blackman looks like there are some [test failures](https://github.com/spinnaker/spin/runs/5424896993?check_suite_focus=true).

robert-blackman:
Hey @dbyron-sf tests are passing locally, is there more here for us to do?

dbyron-sf:
> Hey @dbyron-sf tests are passing locally, is there more here for us to do?

I'm not sure why, but the PR build checks don't seem to be triggering...maybe try pushing an empty commit to see if that wakes things up?

robert-blackman:
> > Hey @dbyron-sf tests are passing locally, is there more here for us to do?
> 
> I'm not sure why, but the PR build checks don't seem to be triggering...maybe try pushing an empty commit to see if that wakes things up?

@dbyron-sf gave it a kick, looks like the workflow needs to be approved

kskewes-sf (approved):

dbyron-sf (approved):

PR #328, opened 2022-03-11
refactor(ci): Correctly pass Version to Docker builds and remove Release Phase var

kskewes-sf:
Preparation for adding new GHA CI for Google Artifact Registry container images as added for other Spinnaker services.

- Consolidate on single `Dockerfile`
- Pass Version and Release Phase through to `go build` so that `spin --version` execution in the final `spin` binary is correct
- Run `go build` directly rather than calling `build.sh` script with the same arguments, then remove script as it is no longer referenced in the repo.

Note: the Jenkins instance which was publishing Spinnaker binaries via `release.sh` is no more so we can do a fresh start of https://spinnaker.io/docs/releases/release-manager-runbook/#one-week-after-branches-are-cut-monday


kskewes-sf:
I would love to rip out `RELEASE_PHASE` entirely. I believe its purpose is covered by `VERSION`.
We can default `VERSION` to dev and move the var into `main.go` so that the ldflag becomes simpler.

If this is amenable please advise and I can either add to this PR or do separately (for 1.28.x).

kskewes-sf:
```
# build with default version
$ go build .
$ ./spin --version
version dev

# override version per Dockerfile
$ go build -ldflags "-X github.com/spinnaker/spin/version.Version=testing" .
$ ./spin --version
version testing
```

dbyron-sf (approved):

PR #340, opened 2022-04-27
chore(swagger): Update master with latest Gate 1.28

kskewes-sf:


kskewes-sf:
It looks like the particular swagger changes are not used by `spin` but updating for posterity.

dbyron-sf (approved):

PR #349, opened 2023-01-10
fix(oauth2): ensure refresh token exists before using expired credential

webmakersteve:
We ran into an issue using an identity provider which did not issue refresh tokens.

Currently the command line tool will fail to authenticate with OAuth2 if using an expired access token and the refresh token is not set. This PR adjusts it to clear the cached credentials if it is expired and there is no refresh token set. The tool will then run the auth routine from the beginning, allowing you to re-log in rather than requiring you edit the spin config file to move forward.


Chinikins:
thanks for fixing this, it was badly needed for awhile :raised_hands:

webmakersteve:
No problem! Already using the binary built off of this commit locally so I could get around the issue myself! :)

webmakersteve:
How do we get this merged into the repo @Chinikins ?

Chinikins (approved):

PR #350, opened 2023-01-10
fix(output): send info strings to stderr

webmakersteve:
Currently the tool sends out informational output to stdout. This makes it hard to pipe output into tools like `jq` and such. 

Adjusts it so only jsonoutput and output go to stdout and the rest should go to stderr via warn.


webmakersteve:
This is technically a breaking change if people depend on the old behavior - so could use some advice on if this is something desired by the team. And if so, how to indicate that in the pr.

webmakersteve:
Anyone available to review this? if it isn't desired it can be closed.

karlskewes:
Thank you for investigating and raising a PR with the potential change.

Is the `quiet` flag insufficient?
```
$ spin | grep quiet
  -q, --quiet                    squelch non-essential output
```

I don't have a Spinnaker setup at the moment to query with spin, but from memory using `--quiet` would trim non-JSON output enabling `jq` or redirect to file usage. I could be mistaken or perhaps it wasn't for all commands..

PR #351, opened 2023-01-12
FIX - When the password contains "$" character the login not working well.

vitorjordao:
When the password contains "$" character in the "config" the login not working well.


cmd/gateclient/client.go
```diff
@@ -235,7 +235,7 @@ func userConfig(gateClient *GatewayClient, configLocation string) error {
 	// but unmarshal to an upstream oauth package, so the cached token needs to match
 	// https://godoc.org/golang.org/x/oauth2#Token
 	if yamlFile != nil {
-		err = yaml.UnmarshalStrict([]byte(os.ExpandEnv(string(yamlFile))), &gateClient.Config)
```
karlskewes, line 238: [os.ExpandEnv](https://pkg.go.dev/os#ExpandEnv) enables use of environment variables in the spin config.
This can be useful for secret injection either locally or in CI.

Unfortunately it doesn't appear possible to escape `$` characters, see: https://github.com/golang/go/issues/43482

In the past I've generated different secrets without the `$` character, is that an option for you?
Alternatively we'd need to look at other solutions though they might be longer or more complicated.

karlskewes: example:
```
$ cat main.go 
package main

import (
	"fmt"
	"os"
)

func main() {
	fmt.Println(os.ExpandEnv("some_key: $my_secret"))
}
```
Then:
```
$ my_secret=my_pass go run main.go
some_key: my_pass
```

karlskewes: To be clear, merging this would break any users of `spin` who use the existing environment variable expansion behaviour/"feature" so I'm a little hesitant to approve this PR.

Appreciate the investigation and thought to raise a PR with a potential option @vitorjordao.

vitorjordao:
I already change my implementation creating a new flag.
This solution will not need to break the old implementation.

PR #355, opened 2023-03-24
fix(pipelineTemplateSave): Take tags from template files

kbatalin:
Spin ignores tags from pipeline template files, and it causes errors during saving. 

Example of the problem:
1. Create a template with a tag "stable" inside
2. Save it via `spin pt save`. Don't pass the tag to the command, because it already exists in the code
3. Front50 parses the template and finds the tag inside: [code](https://github.com/spinnaker/front50/blob/master/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/V2PipelineTemplateController.java#L94)
4. Front50 saves the template with the tag
5. Edit the code of the template and save it again
6. Spin tries to find a template in Front50 without tags (because we don't pass the tag via arguments)
7. Spin can't find the template and invokes "create" method instead of "update"
8. Front50 throws an DuplicateEntityException because a template with the id already exists. Validation code: [code](https://github.com/spinnaker/front50/blob/master/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/V2PipelineTemplateController.java#L115)

Front50 takes a tag from query parameters as a priority. And only if the tag is not in the query parameters, it tries to take it from the template source code. [code](https://github.com/spinnaker/front50/blob/master/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/V2PipelineTemplateController.java#L93) 

I reproduced a similar logic in Spin

### DIFF
--- a/cmd/pipeline-template/save.go
+++ b/cmd/pipeline-template/save.go
@@ -84,6 +84,8 @@ func savePipelineTemplate(cmd *cobra.Command, options *saveOptions) error {
 	getQueryParam := &gate.V2PipelineTemplatesControllerApiGetUsingGET2Opts{}
 	if options.tag != "" {
 		getQueryParam.Tag = optional.NewString(options.tag)
+	} else if tag, exists := templateJson["tag"]; exists && tag.(string) != "" {
+		getQueryParam.Tag = optional.NewString(tag.(string))
 	}
 
 	_, resp, queryErr := options.GateClient.V2PipelineTemplatesControllerApi.GetUsingGET2(options.GateClient.Context, templateId, getQueryParam)
--- a/cmd/pipeline-template/save_test.go
+++ b/cmd/pipeline-template/save_test.go
@@ -387,7 +387,7 @@ func testGatePipelineTemplateUpdateTagSuccess(buffer io.Writer, method *string, 
 			util.NewTestBufferHandlerFunc(http.MethodPost, buffer, http.StatusAccepted, "").ServeHTTP(w, r)
 		}),
 	)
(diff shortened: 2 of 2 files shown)


Chinikins:
did you run `go fmt` on your PR perchance? the lint check is failing, though the error i'm seeing looks like it's more with the job itself than an actual lint error, but figured I would ask anyway

kbatalin:
@Chinikins 
Yes, I ran it for files that I changed. I can run it for the whole project if you are okay with it

karlskewes:
Looks like we need to review & merge #358 first, rebase this and then the lint job should run properly.

karlskewes:
And great work on figuring out the logic and replicating here @kbatalin !

dbyron-sf:
@mergifyio update

kbatalin:
Hi,
@Chinikins @karlskewes @dbyron-sf 

Thanks for the CI fix!
Tests passed

Could you have a look on the PR?

kbatalin:
Hi @Chinikins @karlskewes @dbyron-sf 

Could you have a look on the PR?

kbatalin:
@dbyron-sf 
Yes, sure. Rebased

dbyron-sf:
Yup, hoping @Chinikins or @karlskewes can have a look.

kbatalin:
Hi @Chinikins ,
Could you please have a look?

karlskewes (approved):
LGTM, thanks very much.

kbatalin:
@karlskewes 
Thanks for the approve! Can we merge it now?

dbyron-sf (approved):

PR #358, opened 2023-03-31
chore(lint): Bump GitHub Action versions

karlskewes:
Lint action was failing. Bumping versions for various actions seems to fix it.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,4 +4,4 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "monthly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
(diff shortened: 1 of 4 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
```
karlskewes, line 1: This has worked well for me in other non-spinnaker repositories. 
Be good to add to all of the repositories.

Weekly interval might generate a lot of PR's (especially across all repositories).
See: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#scheduleinterval

.github/workflows/build.yml
```diff
@@ -16,9 +16,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout code
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
+    - uses: actions/setup-go@v4
+      with:
+        go-version-file: go.mod
```
karlskewes, line 22: No need to manually bump the Go version in GHA's anymore. 
`setup-go` can now pull from `go.mod` which is the source of truth.
Works well for me in other non-spinnaker repositories.

dbyron-sf (approved):

PR #366, opened 2023-04-02
Allow polling timeout to be configurable

skandragon:
The old system tried 5 times with an exponential backoff.  The new behavior is based on time, and increases to a maximum of 20 seconds between tries.  It will poll for whatever time is specified on the command line, or in the config file.  Default is 60 seconds, which is the same time the old method would end up using.


karlskewes:
Hey @skandragon, did we miss a review here? Wondering why it was merged...

PR #367, opened 2023-07-26
fix empty list template response

FrancoisPoinsot:
When running `spin pipeline-template list` I was receiving an empty list
This is because the path of the request looked like that
```
/gate/v2/pipelineTemplates?scopes=%26%5B%5D
```

the charactere `%26%5B%5D` decodes to `&[]`
Technically we could say the issue comes from `parameterToString` in client.go, and that function should handle receiving a pointer to an array better.

But since this part is generated, here is a fix on the command layer.
To send an array instead of a pointer to an array


jasonmcintosh (approved):
CLI could seriously use tests, but :) Looks like a good fix in this case ;)

nandak522:
@jasonmcintosh Any idea when is `1.31` release going to happen ?

